PR #4, opened 2018-02-12
add halt and reboot to make clean

mskarbek:
`make clean` doesn't properly clean all build artefacts

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -35,7 +35,7 @@ endif
 
 clean:
 	rm -f *.o *.d
-	rm -f dinit dinitctl shutdown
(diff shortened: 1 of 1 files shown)


davmac314:
Actually those are not build artifacts - they are static shell scripts, not generated by the build process at all. (I know this isn't ideal. For one thing, I'm not sure whether shell scripts are the best option for these or if it would be better to do the traditional "link to the same executable under different names and have it behave differently depending on argv[0]". Also, the shell scripts should probably have the full path to 'shutdown' written in them by the build process).

mskarbek:
From the build/packaging point of view I would prefer symlinks.

PR #15, opened 2019-06-27
Fixed strip on install for cross compilation

fpoussin:
Hi,

This is a small fix for *install -s* to work when cross compiled.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,6 +3,7 @@ include ../mconfig
 HOSTCXX ?= $(CXX)
 HOSTCXXFLAGS ?= $(CXXFLAGS)
 HOSTLDFLAGS ?= $(LDFLAGS)
+STRIP ?= strip
 
 ifeq ($(BUILD_SHUTDOWN),yes)
   SHUTDOWN=shutdown
@@ -39,7 +40,7 @@ dinit: $(dinit_objects)
 
 dinitctl: dinitctl.o
 	$(CXX) -o dinitctl dinitctl.o $(LDFLAGS)
-	
+
 shutdown: shutdown.o
 	$(CXX) -o shutdown shutdown.o $(LDFLAGS)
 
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -54,7 +55,7 @@ run-cppcheck:
 
 install: all
 	install -d $(DESTDIR)$(SBINDIR)
-	install -s dinit dinitctl $(SHUTDOWN) $(DESTDIR)$(SBINDIR)
+	install -s --strip-program=$(STRIP) dinit dinitctl $(SHUTDOWN) $(DESTDIR)$(SBINDIR)
```
vext01: `--strip-program` is not portable I'm afraid. On BSD systems you'd need to set the `STRIP` environment variable instead.

davmac314: I would suggest adding a variable for install instead (ie INSTALL), in the  which can be set to "install -s" or "install -s --strip=..." etc. It could be set by default to "install -s" at the top of the main Makefile via `INSTALL ?= install -s`.

fpoussin: Good idea!

fpoussin: > I would suggest adding a variable for install instead (ie INSTALL), in the which can be set to "install -s" or "install -s --strip=..." etc. It could be set by default to "install -s" at the top of the main Makefile via `INSTALL ?= install -s`.

Turns out it won't work:
```
INSTALL ?= install -s

install -s -d ./install/sbin
install: the strip option may not be used when installing a directory
```

davmac314: Hmm. Maybe just set:

`INSTALL ?= install`

then have:

```
install: all
	$(INSTALL) -d $(DESTDIR)$(SBINDIR)
	$(INSTALL) -s dinit dinitctl $(SHUTDOWN) $(DESTDIR)$(SBINDIR)
```

And you can set:

`INSTALL = install --strip-program=...`

... in your config?

src/Makefile
```diff
@@ -54,7 +55,7 @@ run-cppcheck:
 
 install: all
 	install -d $(DESTDIR)$(SBINDIR)
-	install -s dinit dinitctl $(SHUTDOWN) $(DESTDIR)$(SBINDIR)
+	install $(STRIPOPTS) dinit dinitctl $(SHUTDOWN) $(DESTDIR)$(SBINDIR)
```
fpoussin: This should work everywhere.

src/Makefile
fpoussin: It will work but install will output a warning saying the option has no effect with -d

davmac314:
Perfect. Thanks!

PR #19, opened 2019-08-26
Kernel modules script

fpoussin:
This is a simple script that reads /etc/modules and proceeds with loading each kernel module with its optional arguments.


davmac314:
Looks good, thanks!

PR #34, opened 2021-06-13
Add always-chain option

asherikov:
`always-chain` alters behavior of `chains-to` property forcing the chained service to always start on termination of a service.

### DIFF
--- a/doc/manpages/dinit-service.5.m4
+++ b/doc/manpages/dinit-service.5.m4
@@ -243,10 +243,12 @@ a separate filesystem that is mounted during the first stage; such service
 descriptions will not be found at initial start, and so cannot be started
 directly, but can be chained via this directive.
 
-The chain is not executed if the initial service was explicitly stopped, stopped
-due to a dependency stopping (for any reason), if it will restart (including due
-to a dependent restarting), or if its process terminates abnormally or with an
-exit status indicating an error.
+The chain is not executed if the initial service was explicitly stopped,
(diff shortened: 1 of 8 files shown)


davmac314:
Looks basically fine, I'm just curious about the use case for this (what is the option needed for)?

One minor quibble: the current code uniformly uses `||` and `!` rather than `or` and `not` keywords, please use the same convention.

Also, there is an integration test for `chain-to` (src/igr-tests/chain-to), please modify that test to also test this option, or add a new test.

asherikov:
Fixed style, I'll add test a bit later.
The current goal is to be able to terminate all services and dinit if a particular service stops working for any reason. 
Can also be useful for sending notifications on service termination even if the service is automatically restarted.

asherikov:
Added integration test. 

BTW, a higher level note -- `restart` and `chain-to` are related and can probably be generalized to something like this:
```
on-termination:
    reason: [success|failure|manual]
    service: <this or some other service>
    delay: X.X
    attempts: X
    ...
```

davmac314:
It's true that they could be combined but the use cases are very different. I wouldn't want to have to repeat the service name in a restart directive. Also you can currently specify both restart options and a chain-to service, the latter will only get executed if the service won't restart (e.g if it failed too quickly and too many times); this would get awkward with a common syntax.

davmac314:
(also: merged, thanks!)

PR #42, opened 2022-02-08
Rephrase systemd comparison

Oppen:
The section about shutdown after explaining conflicting services is, IMO, a bit inaccurate.
Implementing shutdown as a service uses the feature of conflicting services, but it is not necessarily the reason for the feature to exist, as there are also some other uses, such as alternative services for the same functionality, say, different network-online providers for different network management services, or classic dbus vs dbus-broker.
So this commit rephrases to avoid misleading interpretations of what the feature is about.

### DIFF
--- a/doc/COMPARISON
+++ b/doc/COMPARISON
@@ -239,11 +239,11 @@ in a more complex package.
 
 The Systemd interdependence graph is more complex than for Dinit and most other
 dependency-handling service managers: a service can conflict with another service, meaning
-that starting one causes the other to stop and vice versa. Systemd implements shutdown
-via a special "shutdown" unit which conflicts with other services so that they stop
-when the shutdown is "started". Other service managers typically do not implement shutdown
-as a service but as a special action; they then don't need to support conflicting
-services.
+that starting one causes the other to stop and vice versa. This allows Systemd to
+implement shutdown via a special "shutdown" unit which conflicts with other services so
+that they stop when the shutdown is "started".
+Other service managers typically need to implement shutdown as a special action because
+they don't support conflicting services.
 
 The "shutdown" unit is just one example of a "special" service. Systemd has several such
 services, for various purposes, including for tight integration with D-Bus (Systemd


davmac314:
> Implementing shutdown as a service uses the feature of conflicting services, but it is not necessarily the reason for the feature to exist

I don't think that's implied by the existing text.

I'm happy for the given change regarding how Systemd handles shutdown:
> This allows Systemd to implement shutdown via a special "shutdown" unit which conflicts with other services so that they stop when the shutdown is "started"

But, the rephrasing of the following sentence changes the nuance and meaning:

I.e. the original text:
> Other service managers typically do not implement shutdown as a service but as a special action; they then don't need to support conflicting services.

The suggested replacement:
> Other service managers typically need to implement shutdown as a special action because they don't support conflicting services

The point of the original text is not that conflicting services allow a service manager to avoid implementing shutdown as a special action; it's that conflicting services are not required to be supported in order to support shutdown (despite that Systemd does use them as part of the shutdown functionality). So, "other service managers typically need to implement shutdown as a special action because they don't support conflicting services" has the wrong nuance (and perhaps isn't correct, see below). Rather, "other service managers can implement shutdown without supporting conflicting services" - which I think the current text conveys.

Note that Systemd also has a special action for shutdown. Systemd documentation:

    halt.target

    A special target unit for shutting down and halting the system. Note that this target is distinct from poweroff.target in that it generally really just halts the system rather than powering it down.

    Applications wanting to halt the system should not start this unit directly, but should instead execute systemctl halt

That last line shows that shutdown is still a special action; you can't (or at least, shouldn't) just start the unit, and support for conflicting services does not appear to have removed the requirement for the service manager to support shutdown as a special action.

Oppen:
> > Implementing shutdown as a service uses the feature of conflicting services, but it is not necessarily the reason for the feature to exist
> 
> I don't think that's implied by the existing text.
> 
I guess it's subjective, but when I read "we don't do X so we don't need to do Y", I read "the reason Y exists is so we can do X".

> I'm happy for the given change regarding how Systemd handles shutdown:
> 
> > This allows Systemd to implement shutdown via a special "shutdown" unit which conflicts with other services so that they stop when the shutdown is "started"
> 
> But, the rephrasing of the following sentence changes the nuance and meaning:
> 
> I.e. the original text:
> 
> > Other service managers typically do not implement shutdown as a service but as a special action; they then don't need to support conflicting services.
> 
> The suggested replacement:
> 
> > Other service managers typically need to implement shutdown as a special action because they don't support conflicting services
> 
> The point of the original text is not that conflicting services allow a service manager to avoid implementing shutdown as a special action; it's that conflicting services are not required to be supported in order to support shutdown (despite that Systemd does use them as part of the shutdown functionality). So, "other service managers typically need to implement shutdown as a special action because they don't support conflicting services" has the wrong nuance (and perhaps isn't correct, see below). Rather, "other service managers can implement shutdown without supporting conflicting services" - which I think the current text conveys.
> 
> Note that Systemd also has a special action for shutdown. Systemd documentation:
> 
> ```
> halt.target
> 
> A special target unit for shutting down and halting the system. Note that this target is distinct from poweroff.target in that it generally really just halts the system rather than powering it down.
> 
> Applications wanting to halt the system should not start this unit directly, but should instead execute systemctl halt
> ```
> 
> That last line shows that shutdown is still a special action; you can't (or at least, shouldn't) just start the unit, and support for conflicting services does not appear to have removed the requirement for the service manager to support shutdown as a special action.

Then something like this may be more appropriate:
> Shutdown generally requires special actions to some degree, so other init systems choose to make the whole process a special action, thus not depending on conflicting services for this process

Does that sound better? What I think is misleading is only that it sounds as the only reason why the feature may be desirable. But yeah, I think my original take sounds a bit like "other systems are hacks" or something like that, which isn't what I intended.

davmac314:
> Does that sound better? What I think is misleading is only that it sounds as the only reason why the feature may be desirable. 

That sounds much better to me, yep!

Oppen:
I never got around rewriting this. I'll close it until I find the time and the words.

PR #44, opened 2022-02-13
use symlinks instead of hardlinks for command aliases, provide "poweroff" alias

q66:
Symlinks are more obvious/easier to track than hardlinks and as far as I can tell there is no disadvantage to using them.

Additionally, provide a "poweroff" command alias. This is to improve compatibility with e.g. `elogind` which internally calls it when you try to shut down the system via `loginctl`.


davmac314:
Merging for now, though probably should be consistent with executable links

davmac314:
*facepalm* the PR does use symbolic links for executables too. Not sure how I missed that. Thanks.

q66:
yes, i have a hook in my distribution's build system that identifies all cases of hardlinks in built packages and errors unless explicitly permitted; this is one of the cases it caught

PR #45, opened 2022-02-13
Also pass CXXOPTS when linking

q66:
It is common convention that compiler flags are passed during linkage, providing increased robustness/convenience when e.g. cross-compiling or in some cases of LTO builds.

### DIFF
--- a/configs/mconfig.Darwin
+++ b/configs/mconfig.Darwin
@@ -11,7 +11,7 @@ SYSCONTROLSOCKET=/var/run/dinitctl
 # Cannot use -fno-rtti: apparently prevents exception handling from working properly.
 CXX=g++
 CXXOPTS=-std=c++11 -Os -Wall -flto
-LDFLAGS=-flto
+LDFLAGS=
 BUILD_SHUTDOWN=no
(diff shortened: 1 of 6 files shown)


davmac314:
Ok in principle, but I would like to remove the redundancy caused by LDFLAGS already having some options that are present in CXXOPTS for Linux in `configs/mconfig.Linux` and `configs/mconfig.Linux.sh` (they already add `-flto -Os` to LDFLAGS).

q66:
ok

by the way, is there any reason to be doing things the way they are done instead of using the conventional CXXFLAGS plus LDFLAGS combo passed on the command line?

davmac314:
Just for simplicity of building. On any of the supported platforms (Linux, FreeBSD, OpenBSD, MacOS) you should be able to get a reasonable build just by typing "make".

You can still put CXXOPTS and LDFLAGS on the command line, this should override the settings from the config files.

I probably should think about renaming CXXOPTS to CXXFLAGS, which is somewhat more (defacto-) standard.

q66:
updated

davmac314:
Merged. Thanks!

Issue #43, opened 2022-02-13
command to export environment variables into activation env


q66:
This would be similar to systemd, which provides a way to export environment variables into its activation environment that is then used for services that are started later. I'm thinking this could perhaps be a command for `dinitctl`; thus it would require the same access level as starting/stopping services and so on.

My main use for this would be able to e.g. smoothly export the dbus session bus address into environment for user services, as well as other session-related variables that are currently not available (because the user-instance of dinit runs separately from the login so that it can properly persist when you say, log in on another tty and then log out on the first one) so that user services can handle things that rely on these, without needing clunky command wrappers.

My other option is to implement this as a part of `dinit-userservd` (which is my daemon that manages automatic user instances of `dinit`) and then provide a special command that will invoke another command through environment managed by `dinit-userservd` and have user services do something like `command = /usr/bin/dinit-run-session another-command ...` but I would first like to hear whether you'd consider implementing this into `dinit` itself (or accept contribution of such functionality) as it would make for nicer service files, and make the functionality available to system services as well (though this could actually be an anti-feature, not quite sure yet)

q66:
This would possibly also be something that would be worth being able to control programmatically. In fact, I think most other service management tasks would also be handy if accessible through some kind of API. Have you considered documenting the control protocol for this?

q66:
OTOH, the more i think about this the more it makes me think that possibly `dinit-userservd` is a better place to do this, since every real use of this i can think of is tied to a login session, so for system services it may not make sense

having to explicitly launch things through a special command can be both good and bad...

davmac314:
I'm ok with this going into `dinit`/`dinitctl`. I was thinking about it already in fact. It's useful for anyone running dinit as a user-level service manager, I think.

PR #46, opened 2022-02-16
command to set a variable in activation environment

q66:
This is a pretty rough draft, mostly to get comments right now.

It adds a new `setenv` command to `dinitctl`. The command can take any number of variables, either as variable names alone (in which case the value is retrieved from the environment `dinitctl` is called in) or in the format `FOO=bar` (which encodes the value).

Limitations:

- length of the name + value + `=` is limited to 1024 minus protocol overhead characters (do we want to bother with a chunked protocol that would let one pass longer vars? i'm leaning towards not, as it seems to be of limited use and likely a sign of doing something terribly wrong)

Fixes https://github.com/davmac314/dinit/issues/43

### DIFF
--- a/src/dinitctl.cc
+++ b/src/dinitctl.cc
@@ -49,6 +49,7 @@ static int add_remove_dependency(int socknum, cpbuffer_t &rbuffer, bool add, con
         const char *service_to, dependency_type dep_type, bool verbose);
 static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, const char *from, const char *to,
         bool enable, bool verbose);
+static int do_setenv(int socknum, cpbuffer_t &rbuffer, char **env_names);
 
 static const char * describeState(bool stopped)
 {
@@ -75,7 +76,8 @@ enum class command_t {
     ADD_DEPENDENCY,
     RM_DEPENDENCY,
     ENABLE_SERVICE,
-    DISABLE_SERVICE
+    DISABLE_SERVICE,
+    SETENV,
 };
 
 class dinit_protocol_error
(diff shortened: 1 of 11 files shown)


src/dinitctl.cc
```diff
@@ -1287,3 +1304,47 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, const char *
...
+            }
+        }
+        uint16_t bufs = buf.size() - 3;
+        // set size in protocol message
+        memcpy(&buf[1], &bufs, 2);
+        // send
+        write_all_x(socknum, buf.data(), bufs + 3);
+        wait_for_reply(rbuffer, socknum);
+        if (rbuffer[0] == DINIT_RP_BADREQ) {
+            if (verbose) {
```
davmac314, line 1339: Don't need `verbose` check for error messages, error output should always visible

davmac314:
Seems fine to me. One niggle is that it would be better to do error checking in dinitctl in case an environment variable does happen to be oversize (> 1024 - 3 bytes) ; it would be better for user experience if `dinitctl` itself gave a specific error message in that case.

Support for chunking larger environment variables would be nice but as you say it's unlikely to be needed in practice.

q66:
alright, i will add the documentation+tests and fix up the code then; i already deployed this in my distribution (to handle session bus address automatically with pipewire user services) and it seems to work quite well

q66:
added integration test, updated manpage, added early length check; if there's anything else let me know, otherwise i think it should be baseline ready

davmac314:
All looks fine.
Only one thing - do you want to add your name to the CONTRIBUTORS file?

q66:
done

PR #48, opened 2022-02-18
some dinitctl socket management enhancements

q66:
This brings two ideas; they are implemented separately, and I am myself not entirely sure if they're good ideas, so I want to raise it here to get some feedback.

1) Both `dinitctl` and `dinit` now read the environment variable `DINIT_SOCKET_PATH` and use it preferentially to determine what the control socket path is. Additionally, `dinit` exports the socket's path into the environment under that name, canonicalized/made absolute. The idea is that when you call `dinitctl` within the service environment, you don't have to worry about passing the socket path (the service may not even know the socket path otherwise, since the user who launches the `--user` instance is in charge of that). However, I'm not sure if both `dinit` and `dinictl` should read the var. Maybe only `dinitctl` should, and `dinit` should only export it.
2) If `XDG_RUNTIME_DIR` is set, it's used instead of `HOME` for the socket location by default (and not hidden anymore). This puts the socket out of the way into a temporary run-time-only location. If not set, the previous behavior is followed.

### DIFF
--- a/doc/manpages/dinit.8.m4
+++ b/doc/manpages/dinit.8.m4
@@ -54,7 +54,8 @@ default is \fI/etc/dinit/environment\fR; see \fBFILES\fR.
 Specifies \fIpath\fP as the path to the control socket used to listen for
 commands from the \fBdinitctl\fR program. The default for the system service
 manager is usually \fI/dev/dinitctl\fR (but can be configured at build time).
-For a user service manager the default is \fI$HOME/.dinitctl\fR.
+For a user service manager the default is either \fI$XDG_RUNTIME_DIR/dinitctl\fR
+or \fI$HOME/.dinitctl\fR, depending on whether \fI$XDG_RUNTIME_DIR\fR is set.
 .TP
 \fB\-l\fR \fIpath\fP, \fB\-\-log\-file\fR \fIpath\fP
 Species \fIpath\fP as the path to the log file, to which Dinit will log status
--- a/doc/manpages/dinitctl.8.m4
+++ b/doc/manpages/dinitctl.8.m4
@@ -77,6 +77,8 @@ the default path to the control socket used to communicate with the \fBdinit\fR 
 .TP
 \fB\-\-socket\-path\fR \fIsocket-path\fR, \fB\-p\fR \fIsocket-path\fR
 Specify the path to the socket used for communicating with the service manager daemon.
+When not specified, the \fIDINIT_SOCKET_PATH\fR environment variable is read, otherwise
+Dinit's default values are used.
 .TP
 \fB\-\-quiet\fR
 Suppress status output, except for errors. 
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -410,10 +410,16 @@ int dinit_main(int argc, char **argv)
     event_loop.init();
 
     if (!am_system_init && !control_socket_path_set) {
-        const char * userhome = service_dir_opt::get_user_home();
-        if (userhome != nullptr) {
-            control_socket_str = userhome;
-            control_socket_str += "/.dinitctl";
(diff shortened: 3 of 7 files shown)


q66:
I guess `dinit` itself should not read `DINIT_SOCKET_PATH` after all, so I will probably change it, but I'd like your feedback for the feature first...

davmac314:
Some thoughts regarding `DINIT_SOCKET_PATH`:

My general rule for dinit has been "avoid hidden/magic parameters unless they provide a clear benefit". Using environment variable values somewhat falls into the category of "hidden parameters".  I guess the first question is, does adding this `DINIT_SOCKET_PATH` variable satisfy a particular use case? I'm guessing it's about being able to run an arbitrary `dinit` session and automatically have the service processes be able to use `dinitctl` to talk to that same session daemon, but is this something you need to do (and is it a big enough need that we should add special support for it, since it could always be accomplished by making the services explicitly use that socket path rather than implicitly use it via the environment variable).

I'm not saying a definitely "no" but I think I'd like more explanation about why it is useful enough to warrant adding it. (I know it's somewhat good for the integration tests but is there a use outside of testing dinit itself? Would there be any situation where a user has a terminal under a `dinit` session that isn't using the default socket path, or is it only a benefit for service scripts/processes - in which case, is it really needed?).

I also think it's better if `dinit` doesn't unconditionally store the value in the environment. Right now dinit doesn't "pollute" the environment namespace in any way (unless told to do so). This could be enabled with a command-line option or just left to whoever/whatever runs `dinit` (since that will need to choose the path anyway, it can easily store it to the environment, unless it uses the default path, in which case the environment variable isn't needed anyway).

I agree that `dinit` probably shouldn't use the value (only `dinitctl`).

Regarding use of `XDG_RUNTIME_DIR`: I'm perfectly happy with this change, on the other hand.

q66:
the reason for storing it in the environment is so that one can run a user instance of dinit (which may have an arbitrary socket path set by the user) while being able to have system-provided service definitions that possibly call dinitctl internally (typically to export something into the dinit environment)

i don't see any other way to do this; the service needs to have some way to be aware of where the socket is, otherwise it can't talk to it

of course, in general just calling dinitctl will work as is because people will generally not change the socket, but if they do, they will be breaking any service file that calls into dinitctl from its command script

q66:
though hm, maybe we could make it so that the variable exists, but only dinitctl reads it, and nothing implicitly exports it? then people running instances of dinit with custom socket who want to use some generic service files can just make sure to export it into the environment their custom dinit is started in, and there should be no problem?

davmac314:
> the reason for storing it in the environment is so that one can run a user instance of dinit (which may have an arbitrary socket path set by the user) while being able to have system-provided service definitions that possibly call dinitctl internally (typically to export something into the dinit environment)

Ok, I get it, although I don't think dinit necessarily has to manage this itself; you could instead have your system services set up to check themselves for `DINIT_SOCKET_PATH` and pass it (via `-p`) to dinitctl -- but, that's a little more effort (and has to be duplicated for each affected service, and would be easy to forget).

> though hm, maybe we could make it so that the variable exists, but only dinitctl reads it, and nothing implicitly exports it? then people running instances of dinit with custom socket who want to use some generic service files can just make sure to export it into the environment their custom dinit is started in, and there should be no problem?

Yes, this seems like the best option to me. I'd also be happy to have a command line option for `dinit` which causes `DINIT_SOCKET_PATH` to be exported automatically (but just as happy without it).

q66:
changed according to your comments

q66:
i'm not quite sure about the option to dinit to automatically export the variable, i'll think about it some more and make it a separate pull request if i come to the conclusion that it's actually good to have (right now i can't think of much of a reason, maybe other than better convenience because you generally want the environment variable to contain an absolute path)

PR #66, opened 2022-06-25
don't mount devpts and shm again

firasuke:
devpts and shm are already being mounted in `early-filesystems.sh`


davmac314:
Merged, thanks!

PR #68, opened 2022-07-11
Add support for dinit readiness notification

q66:
I have not tested this with userservd yet. The mechanism itself works, can be tested e.g. with `dinit --ready-fd 5 5> sock.txt`. By default, it does not report immediately after the socket is open; it will do so just before `dinit` enters the event loop (or after the filesystem is mounted rw), so that services are loaded and scheduled, as realistically you will be waiting for that anyway and this gives more robust guarantees. I will add a testcase once you've confirmed it's okay.


src/dinit.cc
```diff
@@ -266,6 +268,25 @@ static int process_commandline_arg(char **argv, int argc, int &i, options &opts)
...
+        else if (strcmp(argv[i], "--ready-fd") == 0 || strcmp(argv[i], "-F") == 0) {
+            if (++i < argc) {
+                char *endp = nullptr;
+                auto fdn = strtoul(argv[i], &endp, 10);
+                if (endp == argv[i] || *endp) {
+                    cerr << "dinit: '--ready-fd' (-F) requires a numerical argument\n";
+                    return 1;
+                }
+                socket_ready_fd = int(fdn);
+                if (!socket_ready_fd || fcntl(socket_ready_fd, F_GETFD) < 0 || errno == EBADF) {
```
davmac314, line 280: This condition doesn't look right (or am I confused?). I think It should be `&& errno == EBADF` rather than `|| errno == EBADF`, but you could probably skip checking errno altogether in case of failure (EBADF is really the only possibly failure case I think).

q66: hm, no, you're right; i will drop checking the errno

davmac314: Incidentally we should probably set the `CLOEXEC` flag on the file descriptor (using `F_SETFD`) to avoid passing it on to child processes. You could do that here instead of `F_GETFD`.

src/dinit.cc
```diff
@@ -802,13 +830,26 @@ static void control_socket_cb(eventloop_t *loop, int sockfd) noexcept
     }
 }
 
+static void control_socket_ready() noexcept {
+    if (!control_socket_open || socket_ready_fd < 0) {
+        return;
+    }
+    write(socket_ready_fd, control_socket_path, strlen(control_socket_path) + 1);
```
davmac314, line 837: We may as well close `socket_ready_fd` after writing to it here, I think

davmac314:
It looks fine, just a few comments already noted.

src/dinit.cc
q66: the only child processes dinit creates are those of services, right? should those have access to parent file descriptors in the first place? other than the file descriptors used for service readiness notification, shouldn't they all get closed?

src/dinit.cc
q66: generally yes, but that will make things go wrong if you use a standard descriptor (stdout/stderr) to print out the socket

src/dinit.cc
davmac314: They should really - it just doesn't happen yet. It's something I've been meaning to fix. Part of the reason is POSIX doesn't provide a `close_range` call which is the sane way to do such a thing (otherwise you have to loop through all possible file descriptors). However `close_range` is now part of glibc so we could use it on Linux at least (and probably most of the BSDs as well). Up until now dinit has been relying on not inheriting any unwanted file descriptors from the parent process (which is the case when it runs as PID 1, or from the shell, at least), and using CLOEXEC for any other file descriptors it opens, but I agree it's not really correct behaviour.

davmac314: (Adding it to the TODO list now.)

q66: using close_range will make it not work on kernels older than 5.9 (unless runtime detection is used with a fallback)

and a wrapper is still specific to glibc, as far as i can tell musl does not have it (only the raw syscall number)

q66:
reworked things a little to address the comments

davmac314:
Merged, thanks!

Issue #74, opened 2022-07-27
stop-command memory issue


Duncaen:
The argv for stop-command is using a string that is later moved, leaving behind an argv (from `separate_args`) with invalid references to the string.
This specifically happens in my case when the string is <=16 bytes and is stack allocated, apparently any use of the underlying memory would be invalid/undefined after std::move, so just making sure the string is heap allocated is not a full fix.

```
==6138==    at 0x4D03FD3: __execvpe_common (execvpe.c:75)
==6138==    by 0x1257E2: base_process_service::run_child_proc(run_proc_params) (run-child-proc.cc:336)
==6138==    by 0x11ECF8: base_process_service::start_ps_process(std::vector<char const*, std::allocator<char const*> > const&, bool) (baseproc-service.cc:199)
==6138==    by 0x11CB65: scripted_service::bring_down() (proc-service.cc:800)
==6138==    by 0x110C45: service_set::process_queues() (service.h:916)
==6138==    by 0x11BF6D: scripted_service::handle_exit_status(bp_sys::exit_status) (proc-service.cc:614)
==6138==    by 0x11BA47: service_child_watcher::status_change(dasynq::event_loop<dasynq::null_mutex, dasynq::default_traits<dasynq::null_mutex> >&, int, int) (proc-service.cc:238)
==6138==    by 0x11F3E4: dasynq::dprivate::child_proc_watcher_impl<dasynq::event_loop<dasynq::null_mutex, dasynq::default_traits<dasynq::null_mutex> >, service_child_watcher>::dispatch(void*) (dasynq.h:2216)
==6138==    by 0x112323: dasynq::event_loop<dasynq::null_mutex, dasynq::default_traits<dasynq::null_mutex> >::process_events(int) (dasynq.h:1463)
==6138==    by 0x112BA6: dasynq::event_loop<dasynq::null_mutex, dasynq::default_traits<dasynq::null_mutex> >::run(int) (dasynq.h:1501)
==6138==    by 0x10FE5B: dinit_main(int, char**) (dinit.cc:607)
==6138==    by 0x10E609: main (dinit-main.cc:12)
==6138==  Address 0x1ffefff8e0 is on thread 1's stack
==6138==  912 bytes below stack pointer
```

davmac314:
Confirmed, may be an issue for start command also. Will address this soon.

Duncaen:
Yes but only on reload since in the constructor the string is moved first.

davmac314:
[...]

PR #76, opened 2022-07-28
Add support for "Loading any service with `dinit` on `system manager mode`"

### LINKED ISSUE #69: How to load "Specifie" service from bootloader/kernel comand line?
Hello.

 We can start the single service using dinit by entering `single` in the Bootloader command line.
But unfortunately, it is not possible to load another service with this method.
 This is important to me because I'm considering replacing the `single` service with `rescue` and also want to implement a systemd target-like structure in @Dinit-on-Devuan.
 This feature (calling any service using the Bootloader command line) makes it possible to run any target at system startup.

 @davmac314 Now my question is whether it is possible to do this or whether such a feature has been implemented or not?

regards

mobin-2008:
Hi!
This issue is discussed in #69 (although we need to talk more about it now). This time, I decided to add Add support for "Loading any service with dinit on system manager mode" with my c++ knowledge.

Fix #69

## Goal
Add support for loading a service at boot time. its good for create `rescue` service and boot it if necessary also its good for have SystemD target-like services.

## ToDo
- [x] Fix: We can Load a service with `dinit` in `user mode` but `dinit` restricts this feature to `single` service when run in `system manager mode`.
- ~~Pay attention to issue #74 and use `string_ha` instead of `string` if necessary (I have to look into it more).~~ **Its not necessary**
- [x] Modify Documention (in `dinit` man page) (https://github.com/davmac314/dinit/pull/76/commits/fa14434a5363f080a56b35270c05b4afb37ca4c8)
- [x] Add myself to `CONTRIBUTORS` (https://github.com/davmac314/dinit/pull/76/commits/5440558f312a8710dd072e02ab47c4b2da7b8b94)

`Signed-off-by: Mobin <mobin@mobintestserver.ir>`

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -4,4 +4,4 @@ The following people have contributed:
  * Daniel Kolesa - Code, testing, documentation.
  * Edd Barrett - Testing, documentation.
  * Fabien Poussin - Code, services, documentation.
- * Mobin Aydinfar - CI
+ * Mobin Aydinfar - Code, documentation, CI
--- a/doc/manpages/dinit.8.m4
+++ b/doc/manpages/dinit.8.m4
@@ -117,11 +117,16 @@ If none are specified, defaults to \fIboot\fR (which requires that a suitable se
 for the \fIboot\fR service exists).
 .sp
 \fBNote:\fR on Linux, if \fBdinit\fR is running as PID 1 and with UID 0, it will ignore
-service names provided on the command line, other than "single", unless they appear
+service names provided on the command line, other than "single", unless use \fB-t\fR, \fB--service\fR argument or they appear
 anywhere after the "\-o" or "\-m" options (or their long forms).
 This is to filter arguments that were intended for the kernel and not for \fBinit\fR.
 If running in a container, the "\-o" option should be used regardless and will inhibit this
 filtering for any subsequent service names.
+.TP
+\fB\-t\fR, \fB--service\fR \fIservice-name\fR
+Specifies the name of a service that should be started (along with its
+dependencies) and its work when \fBdinit\fR started as \fBinit\fR. you can use this argument(s) in kernel/bootloader command line (E.g: \fB-t tty1\fR).
+also possible to start several services by adding several \fB-t\fR, \fB--service\fR (E.g: \fB-t tty1 -t tty2\fR).
 .\"
 .SH SERVICE DESCRIPTION FILES
 .\"
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -322,10 +322,10 @@ static int process_commandline_arg(char **argv, int argc, int &i, options &opts)
             }
         }
         #endif
-        else if (strcmp(argv[i], "--service") == 0) {
+        else if (strcmp(argv[i], "--service") == 0 || strcmp(argv[i], "-t") == 0) {
             if (++i < argc) {
                 services_to_start.push_back(argv[i]);
-            } 
+            }
             else {
                 cerr << "dinit: '--service' requires an argument\n";
                 return 1;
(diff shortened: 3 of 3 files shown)


davmac314:
> Resolve conflict: The dinit --help says: <service-name> [...] start service with name <service-name>
but its not true! Its only work with single command.

You can specify any `<serivce-name>` if running dinit as a non-system instance. Relevant line is 388:
```
        if (!opts.process_sys_args || strcmp(argv[i], "single") == 0) {
            services_to_start.push_back(argv[i]);
        }
```
if `opts.process_sys_args` is not set, which is the case if pid != 1 or uid != 0, then any name (not just "single") will be treated as a service to start. Otherwise, only "single" is recognised.

mobin-2008:
> > Resolve conflict: The dinit --help says:  [...] start service with name 
> > but its not true! Its only work with single command.
> 
> You can specify any `<serivce-name>` if running dinit as a non-system instance. Relevant line is 388:
> 
> ```
>         if (!opts.process_sys_args || strcmp(argv[i], "single") == 0) {
>             services_to_start.push_back(argv[i]);
>         }
> ```
> 
> if `opts.process_sys_args` is not set, which is the case if pid != 1 or uid != 0, then any name (not just "single") will be treated as a service to start. Otherwise, only "single" is recognised.

So it seems that you are not going to do this right now. no problem
I close this PR.

davmac314:
I'm not sure if you understood me correctly. I was just pointing out that the current help is correct, if dinit is started as a user instance (or at least as a non-system-manager instance). The system-manager behaviour (where it only processes "single" and not other names) is documented in the man page for dinit.

I'd be quite happy to have this change made - and great if you can do it yourself! I'm happy to give whatever guidance/advice I can. Otherwise it will need to wait a little, and I will do it myself.

mobin-2008:
Everything seems right. dinit compiles correctly. Now is the time to test. (Of course, there are still issues such as how to implement and document).

davmac314:
I would suggest a few changes to what you have now:

- you shouldn't need the `specifie_service_loading`/`specifie_service_path` variables (lines 88-89)
- on line 329 and 330, instead of setting those variables, just add the service name directly to the `services_to_start` list at that time (eg `services_to_start.push_back(argv[i])`.
- line 364, please don't remove the original line from the help text as it is still valid to give a service name without prefixing it with `--service`.
- lines 402-404 shouldn't be needed

mobin-2008:
> I would suggest a few changes to what you have now:
> 
>     * you shouldn't need the `specifie_service_loading`/`specifie_service_path` variables (lines 88-89)
> 
>     * on line 329 and 330, instead of setting those variables, just add the service name directly to the `services_to_start` list at that time (eg `services_to_start.push_back(argv[i])`.
> 
>     * line 364, please don't remove the original line from the help text as it is still valid to give a service name without prefixing it with `--service`.
> 
>     * lines 402-404 shouldn't be needed

Done! https://github.com/davmac314/dinit/pull/76/commits/afb3b88755b239912fe1c4443848219e750dd4dc

mobin-2008:
Everything looks right. Now you can start a specific service instead of `boot` & `single` with the `--service` argument. It is also possible to start several services by adding several `--services`:
````
--service tty1 --service tty2
````

Only a few issues remain:
1- Isn't it better to have a distinction between `<service-name>` and `--service <service-name>` in `dinit --help`? It's a bit confusing now.
2- Isn't it better to have a shortcut for it (like `-t`) I myself suggest `-t`, it stands for target, which is similar to what we do with the `--service` argument

davmac314:
> 1- Isn't it better to have a distinction between <service-name> and --service <service-name> in dinit --help? It's a bit confusing now.

I think it's ok as-is, I may amend before release. (It maybe just needs a very small explanation for when you should use `--service <name>` form instead of just `<name>`).

> 2- Isn't it better to have a shortcut for it (like -t) I myself suggest -t, it stands for target, which is similar to what we do with the --service argument

Yes, it would be fine to add that as a shortcut.

mobin-2008:
> I think it's ok as-is, I may amend before release. (It maybe just needs a very small explanation for when you should use `--service <name>` form instead of just `<name>`).

Ok. be I'll leave it as is for now to check it out later.
> Yes, it would be fine to add that as a shortcut.

Ok. i working on it.

mobin-2008:
Everything looks good. Its ready!

davmac314:
Thanks! I'll tweak the wording of the documentation a bit, later, but looks good.

PR #81, opened 2022-08-06
Make Makefiles honor SHUTDOWN_PREFIX properly.

amano-kenji:


davmac314:
Merged. Thanks! I guess this never worked properly (but nobody was using it, or at least no-one reported any issue).

PR #95, opened 2022-08-25
Create mconfig.h for fuzzing

### LINKED ISSUE #91: `make fuzz` target is broken!
Hello.
I was working on the fuzzer when I ran into several problems:
1. `includes` addresses is broken:
https://github.com/davmac314/dinit/blob/ececd169fbfac35035d0ddb6d7cc1d09ae1e5acf/src/tests/cptests/Makefile#L41-L48
(fixed by #92).

2. `prepare-incdir` target not runned by make when building `fuzz` target! its target prepare include dirs for building tests/fuzz (fixed by #94)
3. because `fuzz` target not runned by upstream `MakeFile`, `mconfig.h` not exist but we need `mconfig.h` for buliding `fuzz` target (fixed by #95):
````
❯ make fuzz
clang -O1 -fsanitize=address,undefined,fuzzer-no-link,leak -MMD -MP -Iincludes -I../../../dasynq/include -c ../test-bpsys.cc -o fuzz-test-bpsys.o
clang -O1 -fsanitize=address,undefined,fuzzer-no-link,leak -MMD -MP -Iincludes -I../../../dasynq/include -c ../test-dinit.cc -o fuzz-test-dinit.o
clang -O1 -fsanitize=address,undefined,fuzzer-no-link,leak -MMD -MP -Iincludes -I../../../dasynq/include -c ../../control.cc -o fuzz-control.o
In file included from ../../control.cc:6:
In file included from includes/control.h:16:
In file included from includes/service-listener.h:4:
includes/service-constants.h:4:10: fatal error: 'mconfig.h' file not found
#include "mconfig.h"
         ^~~~~~~~~~~
1 error generated.
````

regards

### PR DESCRIPTION
**Pull request by mobin-2008**
because fuzz target not runned by top level make; we need to create mconfig.h manually.

Fixes #91

### DIFF
--- a/src/tests/cptests/Makefile
+++ b/src/tests/cptests/Makefile
@@ -48,7 +48,7 @@ fuzz_objects = $(foreach obj,$(parent_objs),fuzz-$(obj))
 include/mconfig.h: ../../../build/includes/mconfig.h
 	/usr/bin/env mv ../../../build/includes/mconfig.h includes/mconfig.h
 
-fuzz: prepare-incdir gen-mconfig fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
+fuzz: prepare-incdir include/mconfig.h fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
 	clang++ -std=c++11 -g -O1 -Iincludes -I../../../dasynq/include -fsanitize=fuzzer,address,undefined,leak fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects) -o fuzz
 
 $(fuzz_parent_test_objects): fuzz-%.o: ../%.cc
--- a/src/tests/cptests/Makefile
+++ b/src/tests/cptests/Makefile
@@ -45,7 +45,7 @@ fuzz_objects = $(foreach obj,$(parent_objs),fuzz-$(obj))
 ../../../build/includes/mconfig.h: ../../../mconfig
 	$(MAKE) -C ../../../build all
 
-fuzz: prepare-incdir ../../../build/include/mconfig.h fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
+fuzz: prepare-incdir ../../../build/includes/mconfig.h fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
 	clang++ -std=c++11 -g -O1 -Iincludes -I../../../dasynq/include -I../../../build/includes/ -fsanitize=fuzzer,address,undefined,leak fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects) -o fuzz
 
 $(fuzz_parent_test_objects): fuzz-%.o: ../%.cc
--- a/src/tests/cptests/Makefile
+++ b/src/tests/cptests/Makefile
@@ -45,17 +45,14 @@ fuzz_objects = $(foreach obj,$(parent_objs),fuzz-$(obj))
 ../../../build/includes/mconfig.h: ../../../mconfig
 	$(MAKE) -C ../../../build all
 
-include/mconfig.h: ../../../build/includes/mconfig.h
-	/usr/bin/env mv ../../../build/includes/mconfig.h includes/mconfig.h
-
-fuzz: prepare-incdir include/mconfig.h fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/tests/cptests/Makefile [hunk 1]
```diff
@@ -38,7 +38,13 @@ clean:
 fuzz_parent_test_objects = $(foreach obj,$(notdir $(parent_test_objects)),fuzz-$(obj))
 fuzz_objects = $(foreach obj,$(parent_objs),fuzz-$(obj))
 
-fuzz: prepare-incdir fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
+# Create mconfig.h for fuzzing. because fuzz target not runned by top level make; we need to create mconfig.h manually.
+gen-mconfig:
+	$(MAKE) -C ../../../ mconfig
+	$(MAKE) -C ../../../build/tools/
+	../../../build/tools/mconfig-gen | tee includes/mconfig.h
```
davmac314, line 45: since this rule generates a file (`includes/mconfig.h`) why not make that the target rather than a phony target (`gen-mconfig`)? However, I also favour building mconfig.h in the usual place (`build/includes`) via the existing "make -C ../../../build all" and then just adding the `../../../build/includes` directory to the include path in the compilation targets below

mobin-2008: > since this rule generates a file (includes/mconfig.h) why not make that the target rather than a phony target (gen-mconfig)?

What do you mean? That is, how to make mconfig.h as "target"?

> However, I also favour building mconfig.h in the usual place (build/includes) via the existing "make -C ../../../build all" and then just adding the ../../../build/includes directory to the include path in the compilation targets below

Ok; i use it instead

davmac314: > What do you mean? That is, how to make mconfig.h as "target"?

I mean instead of `gen-mconfig:` you could have `includes/mconfig.h:`. Since this is a real file it means it doesn't get re-built if it doesn't need to. A phony target (`gen-mconfig`) isn't a real file and so gets rebuilt every time.

> Ok; i use it instead

:+1: 

you should probably have something like:

```
../../../build/includes/mconfig.h: ../../../mconfig
	$(MAKE) -C ../../../ mconfig
	$(MAKE) -C ../../../build all
```

mobin-2008: Ok; i fix it now

mobin-2008: We have a new problem.
@davmac314 I dont have any idea about this; Please check this out: https://github.com/mobin-2008/dinit/runs/8034167634

mobin-2008: I run `cat includes/mconfig.h` and everything looks good:
````
// This file is auto-generated by mconfig-gen.cc.

#ifndef DINIT_MCONFIG_H
#define DINIT_MCONFIG_H 1

// Defines

#define SUPPORT_CGROUPS 1

// Constants

constexpr static char DINIT_VERSION[] = "0.16.0pre";
constexpr static char SYSCONTROLSOCKET[] = "/run/dinitctl";
constexpr static char SBINDIR[] = "/sbin";
constexpr static char SHUTDOWN_PREFIX[] = "";

#endif // DINIT_MCONFIG_H
````

Comment (davmac314) [reply src/tests/cptests/Makefile]: Try with latest change on master.

Comment (mobin-2008) [reply src/tests/cptests/Makefile]: Fixed :+1:

File: src/tests/cptests/Makefile [hunk 2]
```diff
@@ -38,7 +38,17 @@ clean:
...
-fuzz: prepare-incdir fuzz.cc $(fuzz_parent_test_objects) $(fuzz_objects)
+# Create mconfig.h for fuzzing. because fuzz target not runned by top level make; we need to create mconfig.h manually.
+../../../mconfig:
+	$(MAKE) -C ../../../ mconfig
+
+../../../build/includes/mconfig.h: ../../../mconfig
+	$(MAKE) -C ../../../build all
+
+include/mconfig.h: ../../../build/includes/mconfig.h
+	/usr/bin/env mv ../../../build/includes/mconfig.h includes/mconfig.h
```
davmac314, line 49: Rather than move the generated mconfig.h, could you just add the ../../../build/includes directory to the include path when compiling (i.e. add `-I../../../build/includes` to the command line for the compilation targets) and change the dependency in `fuzz` from `includes/mconfig.h` to `../../../build/includes/mconfig.h`. Then you can remove this `include/mconfig.h` target altogether.

mobin-2008: Waiting for testing that...

mobin-2008:
Everything looks good!

davmac314 (approved):
Looks good, thanks!

PR #108, opened 2022-09-10
Allow restart of services if all dependents restart automatically

Duncaen:
I have an issue where I need to restart pipewire and its dependents (`pipewire-pulse`, `wireplumber`).

With this change I can restart the `pipewire` service which will stop all the dependents and then restart pipewire which results in the dependents being started again since its made sure that all dependents of pipewire have `restart = true`.

Not sure if the check for `!pinned_started` is correct, but it seems logical to now allow restarts if the dependent is pinned started.

Another way would be to allow `--force` again for restarts, but that leaves the services in a bad state if a dependent doesn't restart on its own.

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -216,7 +216,7 @@ static bool check_restart(service_record *service)
     return true;
 }
 
-bool control_conn_t::check_dependents(service_record *service, bool do_restart, bool &had_dependents)
+bool control_conn_t::check_restart_dependents(service_record *service, bool gentle, bool &had_dependents)
 {
     std::vector<char> reply_pkt;
     size_t num_depts = 0;
@@ -224,7 +224,7 @@ bool control_conn_t::check_dependents(service_record *service, bool do_restart, 
     for (service_dep *dep : service->get_dependents()) {
         if (dep->dep_type == dependency_type::REGULAR && dep->holding_acq) {
             service_record *dep_from = dep->get_from();
-            if (do_restart) {
+            if (!gentle) {
                 // recursively check that no hard dependency is pinned started.
                 if (check_restart(dep_from))
                     continue;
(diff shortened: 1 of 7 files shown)


mobin-2008:
Hi; thanks for PR.
seems like its your first contribution; so Please add your name to `CONTRIBUTORS` file.

davmac314:
I am still trying to figure out what the best way to handle this is. In the PR as it stands there are some things that concern me: for one, it checks only immediate dependents and not their transitive dependents, so requesting a restart might go ahead even if it would result in some service stopping. Also if a transitive dependent does stop it may result in the restart not happening at all.

For example, suppose A (with `restart = no`) "depends-on" B (`restart = yes`) which "depends-on" C, and let's suppose only A is explicitly started (i.e. B and C are started due to being dependencies of A, in the case of C this is a transitive dependency). If a restart for C is issued, this patch would allow it (on the basis that B has `restart = yes`); but it would cause B to stop temporarily, which would cause A to stop (and not restart). Since A would not restart there is now nothing to cause B to be active and so it should not restart, which then also applies to C.

I.e "restart = yes" (`auto_restart` flag) only causes a service to restart if stopping it temporarily doesn't also remove the condition(s) that made it be started in the first place.

Also, `restart = yes` doesn't imply that it's necessarily safe to stop this service without warning. For instance tty login sessions should have `restart = yes` since when you log out another session should begin.

Although it's slightly awkward, one way to deal with this situation might be to have a grouping service which you can use to ensure all the sub-services are started. For example, "pipewire-group" could depend on "pipewire-pulse", "wireplumber" and probably "pipewire". To restart the group of services, first you stop "pipewire" (which causes dependents to stop as well), and then you start "pipewire-group" (which causes dependencies to start). Or, assuming the sub-services are only started due to being a dependency of the group, you can simply `dinitctl restart pipewire-group`.

Duncaen:
Would recursively issuing a restarts for all dependents if `smooth-recovery = false` be better?

I guess then checking `restart = yes` is the wrong in general and should be kept for automatic restarts, what I want is user issued restarts.

Duncaen:
Pushed a new version, that should resolve all the mentioned issues and added a test case.

As basic check, the restart is only allowed when none of the dependents and their dependents is pinned started.

This user requested restart is completely separate from automatic restarts and is done by adding a new variable to the service record that will be used as `with_restart` argument for `do_stop` when `prop_stop` is propagated, which will then propagate the same variable to its dependents and so on.
I also added a new `stopped_reason_t` that indicates a dependency restart, not sure if that is useful/necessary but seemed logical.

davmac314:
Looks ok, but it still has the issue that it allows accidental restarting of services (such as login sessions) without a warning. I think it should require `--force` if any dependents will be restarted. That will require a minor tweak in dinitctl.cc too.

davmac314 (approved):

PR #110, opened 2022-09-12
Meson: Meson build system for dinit!

mobin-2008:
Hi!
## Goal
Meson build system for dinit project!
> Why?

I have thought about it for a while. dinit currently uses make and I think that's enough for now, but the benefits of meson have been tempting me for a while.
Actually, this discussion whether to use meson or make goes back to another discussion (not exactly):
Low-level tools vs. high-level tools:
In the computer world, usually as you move towards the lower levels, the structural limitations that software and programming languages ​​impose on the user become less. In fact (usually) low-level tools leave the user open for different tasks, it can be said that in this case, less work is done by the low-level tool and the user has to do more work. This may seem like a weakness, but I see it as a "trait" because it's not always good or bad, but good or bad depending on the situation. The good thing about this issue is that the user can do his work in the desired way.
On the contrary, it can be said that high-level tools are usually designed to do more work by themselves and put the user's work in a certain framework. This problem makes the user unable to carry out the work according to his wishes due to structural limitations.
In fact, the issue of the user's freedom of action is discussed

long story in short: we need a good balance, do we need make's low-level features? Or is it better to use a build system like Meson?

Advantages of meson:

1. Its clean by default
2. Have build-in support for depends
3. Better compatibility betwen platforms

All of "Advantages of meson" can be done in Make, but there are several problems:
1- It needs a lot of work and time
2- It is usually not optimal (in terms of speed, stability, etc.)

I will say more about it in the future, for now I think its enough.
regards

## ToDo
- [x] Initial `meson.build`
- [x] Initial `meson_options.txt`
- [x] Prepare & create `mconfig.h` in meson
- [x] When host_platform is not  linux; dont build shutdown/reboot/halt
- [x] `supprot_cgroups` option in `meson_options.txt`
- [x] `use_utmpx` option in `meson_options.txt`
- [x] `shutdown_prefix` option in `meson_options.txt`
- [x] Have auto-detecter for `use_new_abi` option
- [x] Have auto-detecter for `cpp_rtti` option
- [x] Have auto-detecter for `-fno-plt` compiler flag
- [x] Have auto-detecter for `supprot_cgroups` option
- [x] Set required meson version in `meson.build` (E.g: `meson_version : '>= 0.53.2'`)
- [x] Unit tests option & target
- [x] Integration tests option & target
- [x] Prepare & install man-pages
- [x] `fuzz` subproject
- [x] Complete docs
- [x] Clean files

`Signed-off-by: Mobin <mobin@mobintestserver.ir>`


mobin-2008:
Why workflows run for "Draft" PR? I must fix it.

meson.build
```diff
@@ -0,0 +1,44 @@
...
+# ToDo. We must verify libstdc++ version and set -D_GLIBCXX_USE_CXX11_ABI to 1 or 0
+elif get_option('use_new_abi').enabled()
+  add_global_arguments('-D_GLIBCXX_USE_CXX11_ABI=1', language : 'cpp')
+elif get_option('use_new_abi').disabled()
+  add_global_arguments('-D_GLIBCXX_USE_CXX11_ABI=0', language : 'cpp')
+endif
+
+## ToDo; its must handeled with `meson_options.txt` instead of force them
+## Use -Wall -flto -fno-rtti -fno-plt
+add_global_arguments('-Wall', language : 'cpp')
```
eli-schwartz, line 35: Wall is simply the default, it's handled by the "warning_level" builtin option.

meson.build
```diff
@@ -0,0 +1,44 @@
...
+add_global_arguments('-flto', language : 'cpp')
```
eli-schwartz, line 36: LTO is handled by the `b_lto` builtin option, so you can just set that to `=true` in default_options.

One of those high-level things. :D

meson.build
```diff
@@ -0,0 +1,44 @@
...
+add_global_arguments('-fno-rtti', language : 'cpp')
```
eli-schwartz, line 37: RTTI is handled by the `cpp_rtti` builtin option, so you can just set that to `=false` in default_options.

eli-schwartz (changes requested):

meson.build
mobin-2008: Hi; Thanks for review :)
Its must added to `default_options` right?

`default_options : ['cpp_std = c++11', 'warning_level = 1'])`

meson.build
mobin-2008: Yes; but i have a question: How can set `cpp_rtti` to auto? i mean its must be disabled in MacOS/FreeBSD & it can be enabled with Linux.
From dinit's BUILD:
https://github.com/davmac314/dinit/blob/b856c50c5a2dd1a8e34e9857603483485f779e65/BUILD#L44-L46

meson.build
mobin-2008: Yep; thanks

davmac314:
To be clear, I don't want to use meson as the default build system for dinit. I prefer that dinit can be built on a minimal POSIX system without additional dependencies (including build tools). In particular meson pulls in Python as a build-time dependency which I would definitely prefer to avoid.

All tools (eg m4) required for building at the moment, other than a C++ compiler, are already required by POSIX.

If you would like to continue this and offer meson build as an alternative, alongside the existing makefiles, that's fine - but please understand that it won't be the default. Also CI should continue to use the default (make). I will personally continue to develop without building using meson.

mobin-2008:
> To be clear, I don't want to use meson as the default build system for dinit. I prefer that dinit can be built on a minimal POSIX system without additional dependencies (including build tools). In particular meson pulls in Python as a build-time dependency which I would definitely prefer to avoid.
> 
> All tools (eg m4) required for building at the moment, other than a C++ compiler, are already required by POSIX.
> 
> If you would like to continue this and offer meson build as an alternative, alongside the existing makefiles, that's fine - but please understand that it won't be the default. Also CI should continue to use the default (make). I will personally continue to develop without building using meson.

I agree. its experimental & we dont want to deprecate Make for one simple reason: "Make" Available on everywhere (Linux; FreeBSD; MacOS & some another UNIX systems) & dont need to "Runtime depends" because its written in C.

eli-schwartz:
> I prefer that dinit can be built on a minimal POSIX system without additional dependencies (including build tools). In particular meson pulls in Python as a build-time dependency which I would definitely prefer to avoid.

This is less of a problem than it sounds, really. https://sr.ht/~lattis/muon is a c99 implementation of Meson, which is a significantly less complex build-time dependency than python (although it's still a tool that isn't preinstalled via posix) and is very easily bootstrapped.

meson.build
eli-schwartz: In this case, the default is already 1 so you needn't bother.

meson.build
eli-schwartz: Is it really mandatory?

Anyway you could do something like this:

```meson
default_options: [
    'cpp_rtti=@0@'.format(host_machine.system() == 'darwin' ? 'false' : 'true'),
]
```

meson.build
mobin-2008: Ok

meson.build
mobin-2008: I think about that

davmac314:
> https://sr.ht/~lattis/muon is a c99 implementation of Meson,

I was aware of muon - it's still an extra dependency though, and it has some limitations (eg it apparently doesn't support cross-compilation, which the Makefiles handle just fine).

mobin-2008:
I add shutdown/reboot/halt building to meson. its have `build_shutdown` option & its can be set to `enabled`, `disabled` & `auto`.
if that option set to `auto`; because dinit dont support shutdown on other POSIX oses (Only Linux supported); if host system not a `Linux` based; meson do not generate shutdown/reboot/halt.
Now there is one problem: can we have something like `SHUTDOWN_PREFIX` in meson?
In addition, several things have been fixed.

mobin-2008:
I search about it; but founds nothing. How to set `#define` in meson? I mean can set some `#define`s as preprocesser argument?
We use a `mconfig.h` file. this file have some `#define`s to set some options. mconfig file genraeted by make & `mconfig-gen.cc` make it to`C`/ `C++` style:
https://github.com/davmac314/dinit/blob/7bee6fe92c8961cbf7ecd7d1452dff1d9890cae2/build/tools/mconfig-gen.cc#L70-L72
We can use this `#define` when we need compile cgroups specifce sections/codes:
https://github.com/davmac314/dinit/blob/7bee6fe92c8961cbf7ecd7d1452dff1d9890cae2/src/dinit.cc#L355-L358

If its possible; we can remove `mconfig.h` in meson builds and use meson options.

eli-schwartz:
https://mesonbuild.com/Configuration.html#configuring-without-an-input-file

mobin-2008:
> https://mesonbuild.com/Configuration.html
> 
> Note: you can do this without an input file, or with an input template.

Yep; i think to do this without original `mconfig`
Thanks :)

mobin-2008:
@davmac314 I add a auto-detecter for `use_new_abi`. Usually `libstdc++` is bundled with `c++ compiler`. This means that they usually have similar versions, since checking the `libstdc++` version is simply not possible, we check the compiler version instead. Just one question, in which version of `clang++` should `use_new_abi` be used/not used?

mobin-2008:
The more I work with `meson`, the more I like it. It provides a good framework for writing everything. I also thought about maintaining `make`: I will maintain `make`. It's actually a good balance: if you have meson, I suggest you use it, meson is faster (I'm a bit unsure about it, especially when standalone `make`-without `autotools` comes to vs `meson`), cleaner and more automated than make, but if you don't use `meson` for whatever reason, no problem, you'll only need a make and dinit (& several POSIX-critical tools like `sh`). I don't intend to consider `make` deprecated, probably after this PR, I will go for several patches for dinit's `make`.
About CI: I wanted ci to test both `meson` and `make` builds, but it seems like it would be expensive, Github actually only allocates 2000 minutes for free per account, and testing `meson` and `make` at the same time is a waste of time. almost doubles the
I will test dinit with Meson myself.

davmac314:
> Just one question, in which version of clang++ should use_new_abi be used/not used?

The problem is in libstdc++, so it depends on the libstdc++ version that it uses (if any), but I don't know if there's any easy way to check that. The problem is gone now even with newer versions of gcc/libstdc++ so you could probably just assume that if clang++ is the compiler then there's no need to define the macro to specify the ABI (especially since clang might be using libc++ instead).

> About CI: I wanted ci to test both meson and make builds

The make build will be the official build and the meson build is an extra contribution that will need to be maintained by yourself (or other external contributors) - so CI shouldn't fail if the meson build gets broken. (I don't want to have to maintain two build systems and other contributors shouldn't need to worry about two build systems).

mobin-2008:
> That doesn't sound right -- the Github Actions limits for public OSS repositories aren't based on minutes, you should get unlimited minutes. Although sometimes you do wait a bit for an available runner -- Meson itself has about 30 CI jobs including a Windows cygwin job that takes anywhere from 1 to 2 hours, 2 hours worth of MSYS2 jobs, an hour worth of macOS jobs, and 12 linux jobs that take between 15 and 30 minutes each.
> 
> And macOS jobs get a 10x multiplier, and Windows ones a 2x multiplier, for the minutes they use.
> 
> Meson doesn't pay for minutes, but any time two PRs, or one push and one PR, is running at the same time, we wait hours for one set of jobs to finish before the next can begin. ;)

Ok; thanks for information. I think about it.
Form docs.github.com:
> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits.

mobin-2008:
> That doesn't sound right -- the Github Actions limits for public OSS repositories aren't based on minutes, you should get unlimited minutes. Although sometimes you do wait a bit for an available runner -- Meson itself has about 30 CI jobs including a Windows cygwin job that takes anywhere from 1 to 2 hours, 2 hours worth of MSYS2 jobs, an hour worth of macOS jobs, and 12 linux jobs that take between 15 and 30 minutes each.
> 
> And macOS jobs get a 10x multiplier, and Windows ones a 2x multiplier, for the minutes they use.
> 
> Meson doesn't pay for minutes, but any time two PRs, or one push and one PR, is running at the same time, we wait hours for one set of jobs to finish before the next can begin. ;)

Ok; thanks for information. I think about it.
Form docs.github.com:
> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits.

meson.build
mobin-2008: I writing `cpp_rtti` as follows:
`'cpp_rtti=@0@'.format(host_machine.system() == 'linux' ? 'true' : 'false'),`
but meson give error: 
`meson.build:2:0: ERROR: Unknown variable "host_machine".`

mobin-2008:
I have already set up target `fuzz`. I'm a bit confused about meson subprojects in docs. My goal is for `fuzz` to be a separate part from the parts of dinit, meaning that `fuzz` is never compiled with other executables. Subproject doesn't seem to offer that.

@eli-schwartz  it seems meson can't set the default compiler. This issue has been brought up before and seems to have been rejected due to "not having a real situation that requires this feature". Now we need fuzz to be compiled with clang/clang++. I know it's not hard to do but it seems nice to have such a feature.

regards

mobin-2008:
seems like on older meson (<0.56): cant set version in `project()`. i figure it out.

mobin-2008:
Its looks very weird but i cant add a value to meson options. We need to add `-lrt` to `cpp_link_args` for FreeBSD uses but i cant add a value to `get_option(cpp_link_args)`:
````
meson.build:63:2: ERROR: Plusassignment target must be an id.
  get_option('cpp_link_args') += '-lrt'
````

I search about it in meson docs & google but finds nothing.

**Comment by eli-schwartz** [comment]
Neither the get_option function nor its return value is a variable pointer, so you cannot assign it a value. And if you could, then it still wouldn't make the coredata see such a change.

If you want to add additional global link arguments, there's a dedicated function for that:

```meson
add_global_link_arguments('-lrt', language: 'c')
```

mobin-2008:
Thanks. Why wasn't this mentioned in the docs?

mobin-2008:
Also i use `add_project_argument` instead of `add_global_argument`. its recommended for some reasons (such as: with `add_global_argument`; project cant be used as subproject).

mobin-2008:
@davmac314 Hi; i have a question: we need to pass `-lrt` to FreeBSD fuzzing?
I think no because original `LDFLAGS` in mconfig not applied on `fuzz` target.

davmac314:
> @davmac314 Hi; i have a question: we need to pass `-lrt` to FreeBSD fuzzing? I think no because original `LDFLAGS` in mconfig not applied on `fuzz` target.

I have never tried to run the fuzzer on FreeBSD, but: if it links fine without `-lrt`, then it doesn't need `-lrt` :)

mobin-2008:
> > @davmac314 Hi; i have a question: we need to pass `-lrt` to FreeBSD fuzzing? I think no because original `LDFLAGS` in mconfig not applied on `fuzz` target.
> 
> I have never tried to run the fuzzer on FreeBSD, but: if it links fine without `-lrt`, then it doesn't need `-lrt` :)

I try it on FreeBSD.

mobin-2008:
@davmac314 seems like; you right! :)

mobin-2008:
I Implement unit tests in meson. everything looks good but `loadtests` fails for no reason. its "Aborted" by system via SIGABRT!
````
==================================== 4/5 =====================================
test:         loadtests
start time:   12:16:10
duration:     0.01s
result:       killed by signal 6 SIGABRT
command:      MALLOC_PERTURB_=212 /home/mobin/dinit_on_meson/github_fork/dinit/dirbuild/loadtests
----------------------------------- stdout -----------------------------------
test_basic...                
----------------------------------- stderr -----------------------------------
terminate called after throwing an instance of 'service_not_found'
````
its ok on make builds but fails on meson. includes or compiler/linker flags is broken? i figure it out.

**Comment by mobin-2008** [comment]
Seems like `-fno-plt` broke Integration tests in FreeBSD 12.3. Dinit crashed with `SIGFAULT`

**Comment by q66** [comment]
> To be clear, I don't want to use meson as the default build system for dinit. I prefer that dinit can be built on a minimal POSIX system without additional dependencies (including build tools). In particular meson pulls in Python as a build-time dependency which I would definitely prefer to avoid.
> 
> All tools (eg m4) required for building at the moment, other than a C++ compiler, are already required by POSIX.

you require GNU make specifically though, which implements tons of poorly specified extensions over POSIX, so that's not exactly clean

File: build/general_variables_for_meson.sh [hunk 4]
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env sh
```
q66, line 1: `/usr/bin/env sh` is pointless cargo culting, `/bin/sh` is *required* by the specs and is the standard guaranteed path (while `/usr/bin/env` is actually not)

meson.build
```diff
@@ -0,0 +1,121 @@
+## Experimetal meson build system for dinit
+project(
+    'dinit', 
+    'cpp', 
+    version : run_command(
+    [find_program('sh'), 'build/general_variables_for_meson.sh', 'version'],
```
q66, line 6: this is terrible, just put the version in there

meson.build
```diff
@@ -0,0 +1,121 @@
...
+    check : true
+    ).stdout(),
+    license : 'Apache',
+    meson_version : '>= 0.47.0',
+    default_options : ['cpp_std = c++11',
+                       'b_lto = true',
```
q66, line 12: drop, same with the prefix

meson.build
```diff
@@ -0,0 +1,121 @@
+## Experimetal meson build system for dinit
+project(
+    'dinit', 
+    'cpp', 
+    version : run_command(
+    [find_program('sh'), 'build/general_variables_for_meson.sh', 'version'],
+    check : true
+    ).stdout(),
+    license : 'Apache',
```
q66, line 9: Apache-2.0

meson.build
```diff
@@ -0,0 +1,121 @@
...
+mandir = get_option('mandir')
+
+## If Meson's default sbindir is modifed; Warn users about it:
+if get_option('sbindir') != 'sbin'
+  warning('Seems like Default "sbindir" is modified. Dinit dont follow Default "sbindir" option. Please use "dinit_sbindir" option instead!')
+endif
+
+## Use -D_GLIBCXX_USE_CXX11_ABI?
+if use_new_abi.auto()
+  if compiler.get_id() == 'gcc'
```
q66, line 46: this is bad, gcc != libstdc++

q66: i think the build system should just not mess with this define at all, it's not portable, impossible to check, and you can just pass it via `CXXFLAGS` from the environment if you care

meson.build
```diff
@@ -0,0 +1,121 @@
...
+elif use_new_abi.enabled()
+  add_project_arguments('-D_GLIBCXX_USE_CXX11_ABI=1', language : 'cpp')
+elif use_new_abi.disabled()
+  add_project_arguments('-D_GLIBCXX_USE_CXX11_ABI=0', language : 'cpp')
+endif
+
+## Use -lrt?
+# We need to pass -lrt to c++ linker on FreeBSD. see BUILD_MESON
+if host_machine.system() == 'freebsd'
+  message('System uses FreeBSD; Pass "-lrt" to linker = YES')
```
q66, line 64: these messages are pointless, and just clutter the output

meson.build
```diff
@@ -0,0 +1,121 @@
...
+  add_project_link_arguments('-lrt', language : 'cpp')
+else
+  message('System not uses FreeBSD; Pass "-lrt" to linker = NO')
+endif
+
+## Use -fno-rtti?
```
q66, line 70: why? this is universal (not tied to linux at all) and meson already has `cpp_rtti` (just put it in `default_options`)

meson.build
```diff
@@ -0,0 +1,121 @@
...
+if host_machine.system() == 'linux'
+  message('System uses linux; Pass "-fno-rtti" to compiler = YES')
+  add_project_arguments('-fno-rtti', language : 'cpp')
+else
+  message('System not uses linux; Pass "-fno-rtti" to compiler = NO')
+endif
+
+## Use -fno-plt?
```
q66, line 78: just drop this whole thing, let people use `CXXFLAGS` if they want it

meson.build
```diff
@@ -0,0 +1,121 @@
...
+if use_utmpx
+  mconfig_data.set('USE_UTMPX', '1')
+else
+  mconfig_data.set('USE_UTMPX', '0')
+endif
+
+## Outputs
+# Standard apps: dinit, dinitctl, dinitcheck, dinit-monitor + shutdown/reboot/halt
+subdir('src')
+# Unit tests
```
q66, line 110: these comments are pointless

q66 (changes requested):
these are just some of the more obvious ones, but otherwise i see lots of other issues with this implementation

i'll point them out when you address these

meson.build
q66: alternatively, just don't mess with rtti and let people who want to disable it manually specify the option (there is no disadvantage to leaving rtti on other than somewhat larger size)

meson.build
mobin-2008: What? i know `prefix` is pointless and must be removed but you marked line 9-12!

q66: i marked the lto line, that's user choice, not a default

mobin-2008: dinit uses a combination of a special configuration file (`mconfig`) and `make` to build itself. `mconfig`s are usually categorized based on the kernel of oses (I think you know that). Each `mconfig` is written to provide the best performance for that target system, in addition to specifying the necessary specifications for that target. For example, all mconfig's `CXXOPTS` contains `-Os` for reduse dinit's execs size by default (if you think thats not good, we can talk about it).
Of course, I may remove `b_lto` in the future, but I will still apply some optimizations by default (and when its necessary, allow users to disable it).

q66: the mconfigs exist because there is no kind of configure script

the choice whether to use LTO or not, same with pretty much all other flags that only affect binary size or performance but not functionality, is entirely user choice and should be left to the user, there is no need to clutter the build system with it

q66: meson has options you can pass at configure time for this reason and the CXXFLAGS environment variable likewise exists for a reason

build/general_variables_for_meson.sh
eli-schwartz: While ISTR /bin/sh exists on every POSIX system, I also STR that the spec doesn't actually say anything about those existing in /bin, merely that they need to exist in whatever path is available as CS_PATH.

mobin-2008: I thought about it. In the past I used `env` to be portable but it seems even `env` is not portable. Some systems put it somewhere other than `/usr/bin` and this will break the script. Unfortunately, there is no solution for this in `POSIX`. `command` is an internal command in the shell and does not have a specific path.
Not worth it at all: I use `#!/bin/sh`. 99% `sh` is exist in `/bin`. Indeed, if you know a system that is not like that, I'm sorry, that system is missing something important!

> Applications should note that the standard PATH to the shell cannot be assumed to be either /bin/sh or /usr/bin/sh, and should be determined by interrogation of the PATH returned by [getconf](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/getconf.html) PATH, ensuring that the returned pathname is an absolute pathname and not a shell built-in.
>
>For example, to determine the location of the standard sh utility:
>
>command -v sh
>
>On some implementations this might return:
>
>/usr/xpg4/bin/sh

Ref: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#tag_20_117_16

mobin-2008: Fixed in https://github.com/davmac314/dinit/pull/110/commits/f6fa6b4af12e4102ed74c0a987d760b8bcda8f0c

meson.build
mobin-2008: The problem is the wrong ABI will cause a crash.
The best case is to have code that doesn't compile/link if user choose the wrong ABI. This is the best case and is also supported in meson.

mobin-2008: see https://github.com/davmac314/dinit/pull/110#issuecomment-1250012202

q66: there will be no crash regardless of the ABI used, because libstdc++ supports both and dinit does not link against any third party libraries where this would matter

this is not something for the build system to concern itself with, especially with extremely fragile autodetection like this, just stick with defaults and if somebody wants something else they can pass it via CXXFLAGS

mobin-2008: but `BUILD` file said:
> Since Dinit is affected by this bug, the unfortunate possibility exists to break Dinit by
upgrading GCC. If you have libstdc++ corresponding to GCC 5.x or 6.x, you *must* build with
the old ABI, but Dinit will be broken if you upgrade to GCC 7. If you have libstdc++ from
GCC 7, you *must* build with the new ABI. If the wrong ABI is used, Dinit may still run
successfully but any attempt to load a non-existing service, for example, will cause Dinit
to crash.

q66: that appears related to some ancient gcc/libstdc++ bug, but considering both gcc 5 and 6 are long EOL, i see no reason to care here (even if somebody did build with gcc5/6, they will not do so with meson, so you are introducing pointless clutter)

mobin-2008: I agree.

mobin-2008:
@q66 Thanks for your review :)
> but otherwise i see lots of other issues with this implementation

What are that problems? For example please just mention that if you like to do it.

q66:
i will re-review this when i get a bit more time

mobin-2008:
Hmmmmmmmmmmmmm, What? Ayo WTH?! why all workflows are stuck at progress? Github action & Cirrus-ci are broken?

doc/manpages/generate-man-pages.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env sh
```
davmac314, line 1: Use `#!/bin/sh`, it's standard and would be consistent

doc/manpages/generate-man-pages.sh
```diff
@@ -0,0 +1,23 @@
...
+
+## Gen man-pages in meson builds
+
+# Why dinit dont use "env" argument in "run_command()"?
+# Because old meson (>0.50.0) dont support "env" in "run_command()".
```
davmac314, line 6: I think you mean `<0.50.0`. But also, why bother supporting old meson versions?

doc/manpages/meson.build
```diff
@@ -0,0 +1,46 @@
...
+    meson.current_build_dir() + 'dinit-monitor.8'
+]
+
+if install_shutdown_man
+    man8 += [
+        meson.current_build_dir() + 'shutdown.8',
+        meson.current_build_dir() + 'reboot.8',
+        meson.current_build_dir() + 'halt.8'
+    ]
+    find_program('ln')
```
davmac314, line 18: I don't think `find_program` does anything if you don't use its return. I'm pretty sure what you would need is something like:

```
ln_cmd = find_program('ln')
run_command(ln_cmd, 'shutdown.8', meson.current_build_dir() + 'reboot.8', check : true)
```
etc.

But, what is the need for `find_program` here anyway? It will just look on the PATH. If you say `run_command('ln', ...)` it will also look on the PATH.

meson.build
```diff
@@ -0,0 +1,74 @@
+## Experimetal meson build system for dinit
+project(
+    'dinit', 
+    'cpp', 
+    version : '0.16.0',
```
davmac314, line 5: Is there any way that you can avoid hardcoding `version` here?

davmac314: (it's better to use from `build/version.conf`. Having it here means this needs to be updated each release).

src/igr-tests/add-rm-dep/run-test.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/sh
 
-../../dinit -d sd -u -p socket -q main &
+if [ $IS_MESON ]; then
+   cd $(dirname $0)
+   DINIT_EXEC=$APPS_PATH/dinit
```
davmac314, line 5: How about just set DINIT_EXEC and DINITCTL_EXEC from outside the script (from the build system, makefile and meson?)

.github/workflows/ubuntu-latest-amd64-fuzzer.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Run fuzzer
       run: src/tests/cptests/fuzz -max_total_time=21000
     - name: Upload crash file(s)
-      uses: actions/upload-artifact@v3.1.0
+      uses: actions/upload-artifact@v3.1.1
```
davmac314, line 24: This change shouldn't be part of this MR

doc/manpages/generate-man-pages.sh
```diff
@@ -0,0 +1,23 @@
...
+MONTH=$4
+YEAR=$5
+
+# Generate man-pages
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/dinit.8.m4 > $TARGET_MAN_PATH/dinit.8
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/dinitctl.8.m4 > $TARGET_MAN_PATH/dinitctl.8
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/dinitcheck.8.m4 > $TARGET_MAN_PATH/dinitcheck.8
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/dinit-monitor.8.m4 > $TARGET_MAN_PATH/dinit-monitor.8
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/dinit-service.5.m4 > $TARGET_MAN_PATH/dinit-service.5
+m4 -DVERSION=$VERSION -DMONTH=$MONTH -DYEAR=$YEAR $SOURCE_MAN_PATH/shutdown.8.m4 > $TARGET_MAN_PATH/shutdown.8
```
davmac314, line 23: This doesn't honour the `shutdown-prefix` option

doc/manpages/meson.build
```diff
@@ -0,0 +1,46 @@
...
+    run_command('ln', '-sf', 'shutdown.8', meson.current_build_dir() + 'reboot.8', check : true)
```
davmac314, line 19: doesn't honour `shutdown-prefix` option

doc/manpages/generate-man-pages.sh
```diff
@@ -0,0 +1,23 @@
...
+# Meson warning:
+# WARNING: Project specifies a minimum meson_version '>= 0.47.0' but uses features which were added in newer versions:
+#  * 0.50.0: {'env arg in run_command'}
+# We use command argument instead of that.
+SOURCE_MAN_PATH=$1
+TARGET_MAN_PATH=$2
+VERSION=$3
+MONTH=$4
+YEAR=$5
+
```
davmac314, line 16: Why not use `set -e` in this script too?

src/igr-tests/meson.build
```diff
@@ -0,0 +1,111 @@
...
+
+## Integration test's Defines
+igr_tests_env = [
+     'IS_MESON=true',
+     'APPS_PATH=' + APPS_PATH
+]
+
+## Prepare integration tests
+run_command([
+    find_program('cp'),
```
davmac314, line 11: No need for `find_program`

src/igr-tests/meson.build
```diff
@@ -0,0 +1,111 @@
...
+    '-r',
+    './',
+    meson.current_build_dir()
+    ],
+    check : true
+)
+
+## Outputs
+# Integration tests
+basic_test_script = find_program('@0@/basic/run-test.sh'.format(meson.current_build_dir()))
```
davmac314, line 21: No need for `find_program` in all these cases, I think

src/tests/cptests/includes/baseproc-sys.h
```diff
@@ -0,0 +1 @@
+../../test-includes/baseproc-sys.h
```
davmac314, line 1: Why has this been added? (and all the others?)

davmac314: The contents of this directory are supposed to be constructed by the build

src/tests/includes/baseproc-sys.h
```diff
@@ -0,0 +1 @@
+../test-includes/baseproc-sys.h
```
davmac314, line 1: No! remove it please

src/tests/meson.build
```diff
@@ -0,0 +1,75 @@
+# Included from top-level meson.build
+
+## Includes
+for_tests_incdir = include_directories(
+    'includes/',
+    '../../dasynq/include/'
+)
+#ToDo; we need to have another meson.build for cptests
```
davmac314, line 8: Is this message still accurate?

davmac314: There is a `meson.build` inside cptests so the comment doesn't seem to make sense

doc/manpages/meson.build
mobin-2008: Seems like you right; I send a commit to fix many issues

doc/manpages/meson.build
mobin-2008: good catch. I'll fix it.

davmac314 (changes requested):
Ok, I have left a bunch of comments. Some are pretty minor.

The main issue I have is the additional of all these symbolic links for the headers in the tests' `include` directories. The makefile build currently copies the various headers there. Admittedly, I'm not very happy with that either and have a better solution in mind which doesn't involve copying/linking headers, but for now I definitely do not want these directories to be populated via checkout and then overwritten as part of the makefile-based build.

I guess it should be possible to have the meson build copy the headers just like the makefile build does?

src/tests/meson.build
mobin-2008: `meson.build` in `cptests` dir is for fuzzer target! I'll send a PR about spliting fuzz target directory and cptests directory. Then I remove this message.

src/igr-tests/meson.build
mobin-2008: Seems like without `find_program`; they don't work!

davmac314:
> The main issue I have is the additional of all these symbolic links for the headers in the tests' `include` directories.

I've thought about this a little more. The main concern isn't the symlinks themselves, it's that the makefile build will overwrite them. One way to address this is to modify the makefile build so it also uses the symlinks. I would be happy with that solution. In this case you should also remove the various `include` directories in the tests from the .gitignore file.

mobin-2008:
> I've thought about this a little more. The main concern isn't the symlinks themselves, it's that the makefile build will overwrite them. One way to address this is to modify the makefile build so it also uses the symlinks. I would be happy with that solution. In this case you should also remove the various `include` directories in the tests from the .gitignore file.

About this issue; Why just don't create a UNIX soft link to needed source files? I mean why create it via build system at build time? Just create some soft symlinks to needed files already!

davmac314:
> About this issue; Why just don't create a UNIX soft link to needed source files? I mean why create it via build system at build time? Just create some symlinks to needed files already!

That is what the last suggestion I made means:
> One way to address this is to modify the makefile build so it also uses the symlinks

i.e. don't overwrite them, just use them as they are.

src/igr-tests/meson.build
davmac314: you can use `files` instead, there is no need to search when you already have the full path

davmac314: eg
```
basic_test_script = files('@0@/basic/run-test.sh'.format(meson.current_build_dir()))
```

meson.build
```diff
@@ -0,0 +1,76 @@
+## Experimetal meson build system for dinit
+project(
+    'dinit', 
+    'cpp', 
+    version : run_command(
+        [find_program('sh'), 'build/general_variables_for_meson.sh', 'version'],
```
davmac314, line 6: it's `/bin/sh`. You've already got it hardcoded in the scripts, why search for it here...

src/igr-tests/meson.build
mobin-2008: thanks. With `files` i can use `meson.current_build_dir()` directly without use `format()`!

src/igr-tests/add-rm-dep/run-test.sh
mobin-2008: I fix this in meson builds but I don't have idea about how can I import some variables to igr-tests in make builds?

davmac314: Make commands are just shell commands...
```
check-igr: igr-runner
	DINIT_EXEC=WHATEVER DINITCTL_EXEC=WHATEVER ./igr-runner
```

src/tests/meson.build
davmac314: hang on, why split fuzz target directory from cptests directory? They are both control protocol tests.
Meson should be able to handle this.

doc/manpages/generate-man-pages.sh
mobin-2008: Because; Newer Mesons don't have a MAJOR feature and don't have something that interest me.
I think it's ok! no?

src/tests/meson.build
mobin-2008: Hmm; Creating a `fuzz` directory in `cptests` dir does not fix our issue? If your your opinion is negative; I'll search for an another way.

doc/manpages/generate-man-pages.sh
davmac314: Doesn't seem ideal if it means we have to sprinkle the source tree with extra shell scripts, but ok

src/tests/meson.build
davmac314: Well, it shouldn't be necessary.  If meson can't handle having two targets in one directory then it's a pretty limited build system (this corresponds to my previous experience with meson, which is part of why I'm resistant to it in the first place).

I don't like either that the current situation has a completely separate build file in cptests which is just for fuzzing. That's wrong. If that can't be fixed in meson, I suggest removing the support for the fuzz target from the meson build altogether. You could do that as a first step anyway and if you can figure out how to get it to work later, raise a second PR then.

(I know the makefile build doesn't currently allow building the fuzzer from the top-level makefile, but it could trivially be made to do so).

eli-schwartz: Meson should not care how many targets you have in a single directory...

doc/manpages/generate-man-pages.sh
eli-schwartz: Maybe I'm totally dumb here, but... why does this need env support at all?

```meson
foreach man: ['dinit.8', 'dinitcheck.8', 'dinitctl.8', 'dinit-monitor.8', 'dinit-service.5', 'shutdown.8']
    section = 'man' + man.split('.')[1]
    custom_target(
        man,
        input: man + '.m4',
        output: man,
        command: [m4, '-DVERSION='+version, '-DMONTH='+month, '-DYEAR='+year, '@INPUT@'],
        capture: true,
        install: true,
        install_dir: join_paths(get_option('mandir'), section),
    )

endforeach
```

Running it as a custom_target has the additional advantage that if you git pull and these man page sources are updated, they actually get rebuilt...

src/tests/meson.build
mobin-2008: I think about it. Seems like my implementation about fuzzing in meson is not fit for our purposes. Also I suggest don't merge this PR until all known issues will fixed.

mobin-2008: I refactored `fuzz` and `cptests` and now they are living in a same directory. @davmac314 That's ok?

mobin-2008:
Meson based CI is started his work!

[builds.sr.ht status](https://builds.sr.ht/~mobin2008/dinit-meson-ci?)

src/tests/meson.build
davmac314: Yes that looks ok, thanks. I have a few other comments still, I'll add separately.

davmac314:
A few general comments/requests:

- can you please rebase and separate into 2 commits: one for changes to the existing build system, and a separate one which adds the meson build
- in shell scripts (igr tests especially), you should always quote expansion of variables which contain file paths (or which otherwise might contain shell special characters). Eg in run-test.sh where there is `$DINIT_EXEC -d sd -u -p socket -q \` it should be `"$DINIT_EXEC" -d sd -u -p socket -q \`. The reason is, if the file path contains a space then `$DINIT_EXEC` will be expanded into two tokens and won't work whereas `"$DINIT_EXEC"` will be correctly expanded to a single token.

src/tests/meson.build
mobin-2008: Ok; I close this

mobin-2008:
> * can you please rebase/squash and separate into 2 commits: one for changes to the existing build system, and a separate one which adds the meson build

Are you sure it's necessary? I mean it's PR have all changes to Add Meson build system to Dinit and can include some fixes to Make. No?

> * in shell scripts (igr tests especially), you should always quote expansion of variables which contain file paths (or which otherwise might contain shell special characters). Eg in run-test.sh where there is `$DINIT_EXEC -d sd -u -p socket -q \` it should be `"$DINIT_EXEC" -d sd -u -p socket -q \`. The reason is, if the file path contains a space then `$DINIT_EXEC` will be expanded into two tokens and won't work whereas `"$DINIT_EXEC"` will be correctly expanded to a single token.

Done in last commit.

davmac314:
> Are you sure it's necessary? I mean it's PR have all changes to Add Meson build system to Dinit and can include some fixes to Make. No?

I would prefer to keep logically separate changes in separate commits. In this case there are two logically separate changes: one is changing the make-based build (so that it's no longer copying headers) and the other one is adding the meson build.

The two commits can still be part of a single PR.

It should be easy to separate them out, but if you have trouble let me know, I'm happy to do it myself.

> 
> > * in shell scripts (igr tests especially), you should always quote expansion of variables which contain file paths [...]
> 
> Done in last commit.

Thanks! I'll go over again soon to make sure it's all looking good.

mobin-2008:
> I would prefer to keep logically separate changes in separate commits. In this case there are two logically separate changes: one is changing the make-based build (so that it's no longer copying headers) and the other one is adding the meson build.
> 
> The two commits can still be part of a single PR.
> 
> It should be easy to separate them out, but if you have trouble let me know, I'm happy to do it myself.

Hmm; Sorry seems like I don't know about what do you think; For now this PR have 129 commits and I don't understand how to have 2 commits to separate something from something else and How that affect to original repo. Seems like I need to learn more about git.

Thanks. If you do it :)

src/igr-tests/add-rm-dep/run-test.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 
-../../dinit -d sd -u -p socket -q main &
+cd $(dirname $0)
```
davmac314, line 3: This should be `cd "$(dirname $0)"`. I.e. there should be quotes around any expansion that is meant to yield a single token but which comes from arbitrary path. If `dirname $0` yields a path with a shell-special character or whitespace, this won't work as is.

src/igr-tests/basic/run-test.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/sh
 
+cd $(dirname $0)
```
davmac314, line 3: quotes needed

src/igr-tests/chain-to/run-test.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/sh
 
+cd $(dirname $0)
```
davmac314, line 3: quotes needed (in the rest of the test too, same issue)

davmac314:
Rebased and merged manually.

PR #125, opened 2022-11-13
Clear errno after trying multiple service dirs

q66:
The previous code would remain errno to be set after trying bad service directories (despite having encountered a valid one later), which would result in nonsensical error messages outside that path.

### DIFF
--- a/src/load-service.cc
+++ b/src/load-service.cc
@@ -296,6 +296,8 @@ service_record * dirload_service_set::load_reload_service(const char *name, serv
             fail_load_path = std::move(service_filename);
         }
     }
+    // We want to clear errno as the service may have been missing in some paths
+    errno = 0;
 
     if (!service_file) {
         if (fail_load_errno == 0) {


q66:
the particular error is `Error reading dependency directory '/whatever' for <whatever> service` in `process_dep_dir`; this is a warning, not a hard error (so it has no effect on behavior), but it still makes no sense for it to say anything (as if you try e.g. 2 service dirs, the first one sets errno and the second one succeeds, we don't care about why the first one failed)

davmac314:
Hmm, I think the better fix location is in `process_dep_dir`. It starts correctly by clearing `errno`:
```
    errno = 0;
    dirent * dent = readdir(depdir);
```
but then in the `while` loop it should be resetting it again just before it does:
```
        dent = readdir(depdir);
```
Fixing it here would be more robust since it would avoid any similar mistakes in the future (generally we shouldn't rely on a non-trivial function call preserving `errno`).

davmac314 (changes requested):
Please set errno to 0 in `process_dep_dir` instead (before the call to `readdir` in the `while` loop).

davmac314:
I'll just make the change myself after merging. Thanks for raising this.

q66:
sorry, i didn't get to checking this again - thanks for addressing it

PR #128, opened 2022-11-21
make: mconfig.Linux.sh: Prefer c++ instead of g++

mobin-2008:
Hi!
On most systems; `c++` is main c++ compiler. For example debian setting `clang++` or `g++` as `c++` exec via `update-alternatives`. so let's follow them!

### DIFF
--- a/configs/mconfig.Linux.sh
+++ b/configs/mconfig.Linux.sh
@@ -35,7 +35,7 @@ test_compile_link_arg() {
   fi
 }
 
-for compiler in g++ clang++ c++ ""; do
+for compiler in c++ clang++ g++ ""; do
   if test -z "$compiler"; then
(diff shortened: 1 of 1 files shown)


mobin-2008:
Hmmm. what? For what; it created in @davmac314's repo? What? Why github created that in your repo? Sorry about that. I don't know github do that by default :/

davmac314:
I deleted the branch from my repo, which seems to have closed this PR. I had forgotten about it, sorry.

I do see the change as unnecessary. What's the benefit? If the user wants to use "c++" as the compiler they can specify it manually. We don't have to cater to whims of various distros that might or might not set "c++" as a meaningful command. The two compilers that we know work are g++ and clang++, that's why they are tried first.

PR #133, opened 2022-12-12
meson: Fix `loadtests` on meson builds

mobin-2008:
Title is enough for commit goal but I have a question about `loadtests` error. When `test-services` directory is not available, `loadtests` print this to `stderr`: `libc++abi: terminating due to uncaught exception of type service_not_found`, This is normal?

`Signed-off-by: Mobin 'hojjat' Aydinfar <mobin@mobintestserver.ir>`

### DIFF
--- a/src/tests/meson.build
+++ b/src/tests/meson.build
@@ -25,6 +25,16 @@ configure_file(
     configuration : mconfig_data
 )
 
+## Prepare loadtests
+run_command([
+    'cp',
+    '-r',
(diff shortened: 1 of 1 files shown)


davmac314:
> When test-services directory is not available, loadtests print this to stderr: libc++abi: terminating due to uncaught exception of type service_not_found, This is normal?

Yes, if the expected test data isn't there, the tests won't work.

davmac314 (approved):

PR #141, opened 2023-01-04
make: Experimental custom `configure` script

mobin-2008:
Hi!
## Goal
a Big overhaul for Build process.
* `configure` script for Dinit
> Why?

I looking for find a good goal for learning more about POSIX sh. Also when I look at old scripts I think they are not ready for future. for example:
* They are not interactable (and Not designed to be interactable! because expected, users use desired arguments with `make` command but it's not a nice idea in packaging dinit for platforms and distros)

so I worked on it. Also `configure` is a very good "goal" for learning more about POSIX shell.
This `configure` script have some features:
* Wrote in Portable POSIX shell (without using something like `autoconf`)
* Universal for all platform (even unsupported platforms) (We can make old scripts "universal" but no one worked on it)
* Interactable (see `--help`)
* Useful logging (Also support `--quiet` mode)
* Checks compiler/linker is working or not and acceptable flags (already `configs/mconfig.Linux.sh` do it) 
* Fast (on my system it's just take 250ms to Done!)

These features makes it a good option to building Dinit rather use old scripts.

> Why do you not contact with me about this thing?

Because My primary reason to do this thing is "Learning more about POSIX shell" so if you reject this PR, It doesn't problem!

## ToDo
- [x] Write `configure`
- [x] Fix some things about function styles (I personally prefer current style but if you don't prefer this style, I change this thing)
- [x] More testing on different cases (and platforms)
- [x] Write docs about new `configure`


configure
```diff
@@ -0,0 +1,360 @@
...
+CXX=$compiler
+CPPFLAGS=$CPPFLAGS
+CXXOPTS=$CXXOPTS
+CXXFLAGS=$CXXFLAGS
+LDFLAGS=$LDFLAGS
+CUSTOM_LDFLAGS=$CUSTOM_LDFLAGS
+BUILD_SHUTDOWN=$BUILD_SHUTDOWN
+SANITIZEOPTS=$SANITIZEROPTS
+
+# Notes:
```
davmac314, line 338: you probably shouldn't generate these comments as they may not be applicable anyway.

src/Makefile
```diff
@@ -21,22 +21,22 @@ all: dinit dinitctl dinitcheck dinit-monitor $(SHUTDOWN)
 $(objects): ../build/includes/mconfig.h
 
 dinit: $(dinit_objects)
-	$(CXX) -o dinit $(dinit_objects) $(CXXOPTS) $(LDFLAGS)
+	$(CXX) -o dinit $(dinit_objects) $(CPPFLAGS) $(CXXOPTS) $(CXXFLAGS) $(LDFLAGS) $(CUSTOM_LDFLAGS)
```
davmac314, line 24: No need for all these changes, see previous comments.

davmac314 (changes requested):
I have made a lot of comments. Another issue is that documentation (instructions in BUILD) should be updated.

If fixed I'm happy to include the script, though not (yet) to have it as default.

configure
```diff
@@ -0,0 +1,360 @@
...
+  --disable-cxxopts             (Not recommanded) Disable ALL CXXOPTS even -std=c++11
+                                  Don't forget set your flags via CXXOPTS & CXXFLAGS envs otherwise
+                                  build may be failed!
+  --disable-ldflags             (Not recommanded) Disable ALL LDFLAGS even "-lrt on FreeBSD"
+                                  You can use LDFLAGS and CUSTOM_LDFLAGS envs to set your ld flags.
+
+Environment variables can be used:
+  CXX                           If you wanna use specific C++ compiler.
+  CPPFLAGS                      If you wanna use some arguments in C++ compiler command line.
+                                  It's passed before CXXOPTS
```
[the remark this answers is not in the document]
mobin-2008: I'll fix this thing.
> Also, CPPFLAGS are traditionally preprocessor flags. If this is just intended as a way to supply extra flags then it should probably be named something else. But also, is it necessary at all?

Also, I think yes, because sometimes you need to set a compiler flags before `CXXOPTS` (For example seems like Debian for "Hardening" do this thing).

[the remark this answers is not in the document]
davmac314: Can you give me an example of an option that needs to go before the other options? Typically later options override earlier ones so there's no advantage to having options go earlier on the command line

configure
```diff
@@ -0,0 +1,360 @@
...
+Usage: $0 [OPTION]...
+
+Defaults for the options are specified in brackets.
+
+  --help                        This help message.
+  --quiet                       Don't print normal messages, just errors.
+  --clean                       Clear mconfig and configure's temp files.
+
+Target options:
+  --target                      Set the target system (for cross-compiling) []
```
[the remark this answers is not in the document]
mobin-2008: It's from `autotools` (specially `autoconf`) legacy. I writed this for hold compatibility with `autoconf` generated `configure`s.
See: https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html
Use it is simple:
```
$ configure --target=aarch64-linux-gnu --build=x86_64-linux-gnu
```
or
```
$ configure --host=i686-linxu-gnu --build=x86_64-linux-gnu
```
or
```
$ configure --target=aarch64-linxu-gnu
```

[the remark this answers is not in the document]
davmac314: For autotools it is potentially used for more than just selecting the compiler/build tools chain. But anyway I don't think we should be concerned about autotools compatibility - we're not setting out to remake autotools.

For example, it doesn't make sense that there is both `--target` and `--platform`. What if I specify:

`--target=i686-linux-gnu --platform=FreeBSD`?

Currently it is possible to specify compiler, target, and platform - but target is meaningless if compiler is specified, so I think it's better not to have it. Otherwise it's not clear if I specify `--target` do I also need to specify `--platform` for example? (I mean: I understand what they do. But for other users, it will not be clear).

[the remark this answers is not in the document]
mobin-2008: `--platform` just for don't set some arguments to compiler/linker. So "For cross-compiling just set CXX and HOSTCXX options" right?

configure
```diff
@@ -0,0 +1,344 @@
...
+fi
+
+## Create mconfig
+rm -f testfile*
+info Creating mconfig...
+cat << EOF > mconfig
+## Auto-generated by "$0" for "$PLATFORM"
+# Installation path options.
+
+SBINDIR=$SBINDIR
```
davmac314, line 312: Note that this will fail to correctly quote the output. Eg try `--sbindir="/usr/some dir"` or some other path containing a space. There's no easy way to solve that though.

configure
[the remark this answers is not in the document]
davmac314: Well selecting the platform also selects suitable defaults for some of the other options. So I would say for cross compiling you should specify CXX, HOSTCXX, and `--platform`.

configure
mobin-2008: I'm think it's easiest solution: "Please use `\` (E.g `/usr/some\ dir`)  for paths containing a space"

davmac314: yeah but if you do `./configure --sbindir="/usr/some\ dir" it won't work. The user would need to double the backslash: `./configure --sbindir="/usr/some\\ dir". Don't worry about it for now.

mobin-2008: Are you sure about this, I do `./configure --sbindir="/usr/some\ dir"` and it works.
> Don't worry about it for now.
Ok

davmac314: > Are you sure about this,
Actually, no, you are right. I thought that `\x` for any x would be collapsed to just `x` but this is not the case apparently.

configure
```diff
@@ -0,0 +1,360 @@
+#!/bin/sh
+### Custom configure script of Dinit.
+
```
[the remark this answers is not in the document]
mobin-2008: Hm. about `set -e`, I created a error system as `error()` and `sub_error()` function, so when we catch a error, print a error message and exit immediately. (except before `error()` and `sub_error()` ) so I think I don't use this thing.
About `set -u`, My script checks a variable exist or not, `set -u` broke this thing.
```
Info: Checking C++ Compiler...
./configure: 223: CXX: parameter not set
```
but `$CXX` is not defined in some cases. so if we need to use `set -u`, I need to change some basic things.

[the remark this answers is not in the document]
davmac314: `set -e` is good practice for catching cases where you forgot to explicitly check. I can see that you have put explicit checks for a lot of things and these will continue to work even if you use `set -e`.

If using `set -u` and you want to refer to a variable that may not be set, you can use `${VARNAME:-}`.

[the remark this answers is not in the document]
mobin-2008: I do it. but seems like FreeBSD CIs are fail. I figure it out.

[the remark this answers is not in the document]
mobin-2008: FreeBSD CI has fixed.

configure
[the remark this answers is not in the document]
mobin-2008: Seems like you right.

davmac314:
I have unsubscribed from notifications (I get a notification for every commit/merge otherwise). You might need to message me some other way when you need a final review. (In future it might be better to avoid pushing so many small changes - one typical workflow is to only push once you need it reviewed, and squash into a few commits when you do).

mobin-2008:
> I have unsubscribed from notifications (I get a notification for every commit/merge otherwise). You might need to message me some other way when you need a final review. (In future it might be better to avoid pushing so many small changes - one typical workflow is to only push once you need it reviewed, and squash into a few commits when you do).

For every commit? :/ that's bad.
Ok, When I need to your final review, I ping you in this PR.

mobin-2008:
@davmac314 I think we just need to mention `configure` and just give some special notes about that. We don't need to mention every option. No?

davmac314:
> @davmac314 I think we just need to mention `configure` and just give some special notes about that. We don't need to mention every option. No?

Right, no need to mention every option, just point out that you can use `./configure --help`.

mobin-2008:
@davmac314 I think it's ready!

BUILD
```diff
@@ -144,6 +150,25 @@ The dinit executable will be put in /sbin (or rather, in $DESTDIR/sbin), which m
...
 
+Special notes for "configure" script
+=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+
+When you using experimental "configure" script, be careful about this things:
+
+Seting environment variables in "configure" arguments: This thing works but need a backslash (\)
+with double quotation (") for every space. For example:
+
+    ./configure CXXOPTS_POST="-std=c++14\ -O3\ -g"
```
davmac314, line 161: For file paths I can accept that spaces need special treatment but it shouldn't be necessary for options like this. Please find a robust way to make this work. I think the issue currently is the `eval` on line 215; you can't just use a simple `eval` on input supplied by the user, so many things can go wrong.

configure
```diff
@@ -0,0 +1,338 @@
+#!/bin/sh
+### Custom configure script of Dinit.
+## Initial prepartion
+set -eu
+cd "$(dirname "$0")"
+. build/version.conf
```
davmac314, line 6: Don't include a makefile fragment into a shell script. I don't think it's needed anyway?

configure
```diff
@@ -0,0 +1,338 @@
...
+## Helper funcations
+# See http://www.etalabs.net/sh_tricks.html
+echo()
+{
+    IFS=" " printf %s\\n "$*"
+}
+
+info()
+{
+    echo "Info: $*"
```
davmac314, line 17: Printing "Info: " before everything is superfluous.

configure
```diff
@@ -0,0 +1,338 @@
...
+fi
+cxx_works "$compiler"
+if [ -n "${hostcompiler:-}" ]; then
+    cxx_works "$hostcompiler"
+fi
+
+## Find depends
+if [ "$PLATFORM" = "Linux" ] || [ "$PLATFORM" = "Darwin" ]; then
+    info Checking make is executable...
+    find_program make
```
davmac314, line 267: why are you trying to find 'make' (or 'gmake')?

BUILD
```diff
@@ -9,9 +9,15 @@ suitable build configuration is provided and will be used automatically if no ma
 is supplied - skip directly to running "make" (more details below) if you are on one of these
 systems and are happy to use the default configuration.
 
-For other systems, or to fine tune or correct the configuration, create and edit the
-"mconfig" file (start by copying one for a particular OS from the "configs" directory) to choose
-appropriate values for the configuration variables defined within. In particular:
+For other systems, or to fine tune or correct the configuration, you can use experimental
+"configure" script. This script try to find C++ compiler and auto-detects OS and checks some options
+are supported or not. Also this script have a interface to working with them, for more info use
```
davmac314, line 14: " Also this script have a interface to working with them" phrasing is very awkward, you could just say "several options are available to control  configuration" for example

BUILD
```diff
@@ -9,9 +9,15 @@ suitable build configuration is provided and will be used automatically if no ma
...
+
+    ./configure --help
+
+command to see all options.
+otherwise you can create and edit the "mconfig" file (start by copying one for a particular OS
+from the "configs" directory) to choose appropriate values for the configuration variables defined within. In particular:
```
davmac314, line 20: This line has is too long, the file generally has lines < 100 characters in length

configure
```diff
@@ -0,0 +1,338 @@
+#!/bin/sh
+### Custom configure script of Dinit.
+## Initial prepartion
+set -eu
+cd "$(dirname "$0")"
+. build/version.conf
+
+## Helper funcations
+# See http://www.etalabs.net/sh_tricks.html
```
davmac314, line 9: Don't just refer to a link (which may become invalid), include the reason in the comment.

configure
```diff
@@ -0,0 +1,338 @@
...
+                                  Don't forget set your flags via CXXOPTS_POST env otherwise
+                                  build may be failed!
+  --disable-ldflags             (Not recommanded) Disable ALL LDFLAGS even "-lrt on FreeBSD"
+                                  You can use LDFLAGS_POST env to set your ld flags.
+
+Environment variables can be used:
+  Note: CXXOPTS and LDFLAGS are Default compiler/linker flags. Don't use them. because
+  All changes will be overwrited.
+
+  CXX                           If you wanna use specific C++ compiler.
```
davmac314, line 157: "wanna" should be "want to"

configure
```diff
@@ -0,0 +1,338 @@
...
+  --enable-utmpx                Enable manipulating the utmp/utmpx database via the related POSIX functions [auto]
+  --disable-utmpx               Disable manipulating the utmp/utmpx database via the related POSIX functions
+  --disable-cxxopts             (Not recommanded) Disable ALL CXXOPTS even -std=c++11
+                                  Don't forget set your flags via CXXOPTS_POST env otherwise
+                                  build may be failed!
+  --disable-ldflags             (Not recommanded) Disable ALL LDFLAGS even "-lrt on FreeBSD"
+                                  You can use LDFLAGS_POST env to set your ld flags.
+
+Environment variables can be used:
+  Note: CXXOPTS and LDFLAGS are Default compiler/linker flags. Don't use them. because
```
davmac314, line 154: This is confusing still. I think you should either remove all reference to CXXOPTS and LDFLAGS from the output, or allow them to be overridden just like other variables (and then you can get rid of the confusing `--disable-cxxopts`/`--disable-ldflags` options).

Saying they are "default compiler/linker flags" is ambiguous. It's better to be specific: they will be set to the options that the configure script uses as default for the platform, filtered to remove any options not supported by the compiler.

configure
```diff
@@ -0,0 +1,338 @@
...
+        --enable-cgroups) SUPPORT_CGROUPS=1 ;;
+        --disable-cgroups) SUPPORT_CGROUPS=0 ;;
+        --enable-sanitize) NO_SANITIZE=0 ;;
+        --disable-sanitize) NO_SANITIZE=1 ;;
+        --enable-utmpx) USE_UTMPX=1 ;;
+        --disable-utmpx) USE_UTMPX=0 ;;
+        --disable-cxxopts) NO_CXXOPTS=true ;;
+        --disable-ldflags) NO_LDFLAGS=true ;;
+        -*) warning Unknown argument: "$arg" ;;
+        *=*) eval "$arg" ;;
```
davmac314, line 215: This is the problematic `eval`

configure
davmac314: I.e. make must be there to build. It doesn't matter what it is called; the user will have to invoke it anyway. I don't think you should fail configure just because you can't find the make executable. The user should know where it is.

configure
```diff
@@ -0,0 +1,338 @@
...
+        sub_info "$(command -v "$1")"
+    else
+        sub_error "$1" is not executable/exist!
+    fi
+}
+
+try_cxx_argument()
+{
+    info Checking whatever compiler accpets "$2"...
+    if $compiler -c $CXXOPTS $2 ${POST_CXXOPTS:-} testfile.cc -o testfile.o > /dev/null 2>&1; then
```
davmac314, line 89: The makefiles currently use CXXOPTS for compilation _and_ linking. So you need to make sure it works for the options work for both.

davmac314:
Ok, I've added a few more comments, I hope they make sense to you.

BUILD
```diff
@@ -9,9 +9,15 @@ suitable build configuration is provided and will be used automatically if no ma
 is supplied - skip directly to running "make" (more details below) if you are on one of these
 systems and are happy to use the default configuration.
 
-For other systems, or to fine tune or correct the configuration, create and edit the
```
davmac314, line 12: Don't remove the "edit mconfig" option in favour of running the script. Fine tuning or correcting the configuration is still done by editing mconfig, but one way of editing mconfig is running the script. I.e. it should be clear what running the configure script does: it generates an mconfig for you. You should add the information about the configure script after the existing text, not replace the existing text.

BUILD
```diff
@@ -144,6 +150,25 @@ The dinit executable will be put in /sbin (or rather, in $DESTDIR/sbin), which m
 path for normal users. Consider making a symbolic link to /usr/sbin/dinit.
 
 
+Special notes for "configure" script
```
davmac314, line 153: Put these in a separate file, or in the `--help` output from the configure script.

davmac314:
Just added a couple of extra comments, sorry. That should be all.

BUILD
davmac314: Actually - you have changed "make" so that it generates a configuration using the configure script if the platform isn't recognised, so the documentation changes here need to reflect that too.

mobin-2008: I do first part but I need to think about second part.

configure
```diff
@@ -95,7 +95,7 @@ find_program()
 try_cxx_argument()
 {
     info Checking whatever compiler accpets "$2"...
-    if $compiler -c $CXXOPTS $2 ${POST_CXXOPTS:-} testfile.cc -o testfile.o > /dev/null 2>&1; then
+    if $compiler -c $CXXOPTS $2 ${POST_CXXOPTS:-} $LDFLAGS ${POST_LDFLAGS:-} testfile.cc -o testfile.o > /dev/null 2>&1; then
```
davmac314, line 88: It's not that you need to add $LDFLAGS to the compilation test, it's that the compilation test should link as well as compile. So you need to remove `-c` and change `-o -testfile.o` to just `-o testfile`. See mconfig.Linux.sh file for example.

davmac314: (Though, it does make sense to include `${POST_LDFLAGS:-}` since we want to check that any automatically chosen option is compatible with user-supplied link flags - so that's a good idea!)

mobin-2008: Hold on a second, `${POST_LDFLAGS:-}` or `${LDFLAGS_POST:-}`? seems like I need to fix this thing.

davmac314: They should be named consistently. If it's CXXOPTS_POST then it should be LDFLAGS_POST!

configure
mobin-2008: I think if users need to avoid those arguments, just pass a `CXXOPTS` ( or `LDFLAGS`) var with empty value.
```
$ ./configure CXXOPTS=""
```
I working on it. but with `set -u` it is more difficult to do.

mobin-2008: I do this thing. Now users can replace `CXXOPTS` and `LDFLAGS` without any "confiusing option". Also if users want to fully disable `CXXOPTS` or `LDFLAGS` just pass `CXXOPTS=""` in arguments or pass `CXXOPTS="null"` as environment variable.
It's ok?

mobin-2008:
@davmac314 a little off-topic: Are you sure 0.17.0pre is "Alpha release (#7)"? Is not "Beta release (#1)"?

davmac314:
> @davmac314 a little off-topic: in `README.md` are you sure 0.17.0pre is "Alpha release (#9)"? Is not "Beta release (#1)"?

That needs to be decided when the release is drafted. Currently I'm hoping that 0.17.0 will be the first beta release. But, I won't change the description in the README in case it needs to be changed back. Better to update it when the release is ready and status is known.

davmac314:
Note also that `0.17.0pre` is definitely not the beta release (0.17.0 may be beta, 0.17.0pre is by definition pre-beta), so it would be wrong to change it say beta now.

mobin-2008:
> Note also that `0.17.0pre` is definitely not the beta release (0.17.0 may be beta, 0.17.0pre is by definition pre-beta), so it would be wrong to change it say beta now.

I think it's project "target". so, yeah you right.

configure
mobin-2008: I mention this thing in `./configure --help`

configure
davmac314: > or pass CXXOPTS="null" as environment variable.

Why not just `CXXOPTS=""` regardless of how it is passed (environment or argument)?

mobin-2008: because I don't know how we can detect `CXXOPTS` is user defined variable (and have empty value) or it's just not exist! Do you have a solution for this?

davmac314: Try searching for shell parameter expansions rules.

`${VARIABLE+IS_SET}` will expand to `IS_SET` if the variable is set (including if set to ""), or the empty string ("") otherwise, for example.

mobin-2008:
I think some options to modify some default compiler flags are useful:
```
  --enable-lto                  Enable link-time optimisation [Enabled]
  --disable-lto                 Disable link-time optimisation
  --enable-rtti                 Enable C++ Run-time type information, Dinit doesn't need that [Enabled only on MacOS]
  --disable-rtti                Disable C++ Run-time type information
  --enable-minsize              Enable -Os flag [Enabled]
  --disable-minsize             Disable -Os flag
```
@davmac314, I'm not sure about adding these things to `configure`, What do you think?

davmac314:
I'm not sure, you can already set CXXOPTS and CXXOPTS_POST, now if you have `--enable-minsize` for example does it conflict with those or ...?

I don't think there should be too many ways to do the same thing, especially if they can conflict with each other. I'm not totally opposed to these options as long as the documentation is clear though.

More general question: is this almost ready? I reviewed quite a while ago but you've done a lot of commits since then. It would be good to close off the MR, if you want to do incremental changes afterwards you can always open another MR. (Also: please squash the commits!)

mobin-2008:
> I'm not sure, you can already set CXXOPTS and CXXOPTS_POST, now if you have `--enable-minsize` for example does it conflict with those or ...?
> 
> I don't think there should be too many ways to do the same thing, especially if they can conflict with each other. I'm not totally opposed to these options as long as the documentation is clear though.

Seems like it's very complex idea, I don't add those thing to `configure` (at least for yet).
> More general question: is this almost ready? I reviewed quite a while ago but you've done a lot of commits since then. It would be good to close off the MR, if you want to do incremental changes afterwards you can always open another MR. 

Yes it's ready! I just waited for you because you said about some changes to Makefiles then those are maybe affect `configure`. otherwise it's ready to be merged.

> (Also: please squash the commits!)

Just use Github build-in for squash these commits. This PR just include a script and some related documents in `BUILD`.
For later, I squash all commits in PR with two branches (a branch and `_wip` branch).

Regards

davmac314:
> Yes it's ready! I just waited for you because you said about some changes to Makefiles then those are maybe affect `configure`

Ah right, I had forgotten that I had mentioned that. I've pushed the changes I wanted now. Unfortunately it might cause a merge conflict with BUILD file.

The variables expected to be set in mconfig are a bit different now, typical example is:
```
CXX=g++
CXXFLAGS=-std=c++11 -Os -Wall -flto -fno-rtti -fno-plt
CPPFLAGS=-D_GLIBCXX_USE_CXX11_ABI=1
LDFLAGS=$(CXXFLAGS)
LDFLAGS=$(CXXFLAGS)
TEST_CXXFLAGS=$(CXXFLAGS) -fsanitize=address,undefined
TEST_LDFLAGS=$(LDFLAGS) $(TEST_CXXFLAGS)
BUILD_SHUTDOWN=yes
```

Note SANITIZEOPTS is gone, CXXOPTS is renamed to CXXFLAGS, CPPFLAGS is new (for preprocessor options), TEST_CXXFLAGS and TEST_LDFLAGS are new options for compiling/linking tests. It would be good if you can make your script generate these.

mobin-2008:
> > Yes it's ready! I just waited for you because you said about some changes to Makefiles then those are maybe affect `configure`
> 
> Ah right, I had forgotten that I had mentioned that. I've pushed the changes I wanted now. Unfortunately it might cause a merge conflict with BUILD file.
> 
> The variables expected to be set in mconfig are a bit different now, typical example is:
> 
> ```
> CXX=g++
> CXXFLAGS=-std=c++11 -Os -Wall -flto -fno-rtti -fno-plt
> CPPFLAGS=-D_GLIBCXX_USE_CXX11_ABI=1
> LDFLAGS=$(CXXFLAGS)
> TEST_CXXFLAGS=$(CXXFLAGS) -fsanitize=address,undefined
> TEST_LDFLAGS=$(LDFLAGS) $(TEST_CXXFLAGS)
> BUILD_SHUTDOWN=yes
> ```
> 
> Note SANITIZEOPTS is gone, CXXOPTS is renamed to CXXFLAGS, CPPFLAGS is new (for preprocessor options), TEST_CXXFLAGS and TEST_LDFLAGS are new options for compiling/linking tests. It would be good if you can make your script generate these.

I think it's ready.

configure
```diff
@@ -0,0 +1,363 @@
...
+        --enable-shutdown|--enable-shutdown=yes) BUILD_SHUTDOWN=yes ;;
+        --disable-shutdown|--enable-shutdown=no) BUILD_SHUTDOWN=no ;;
+        --enable-cgroups|--enable-cgroups=yes) SUPPORT_CGROUPS=1 ;;
+        --disable-cgroups|--enable-cgroups=no) SUPPORT_CGROUPS=0 ;;
+        --enable-sanitize|--enable-sanitize=yes) NO_SANITIZE=0 ;;
+        --disable-sanitize|--enable-sanitize=no) NO_SANITIZE=1 ;;
+        --enable-utmpx|--enable-utmpx=yes) USE_UTMPX=1 ;;
+        --disable-utmpx|--enable-utmpx=no) USE_UTMPX=0 ;;
+        *=*)
+            envvar=$(expr "x$arg" : 'x\([^=]*\)=')
```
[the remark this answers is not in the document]
davmac314: Also maybe have a special case to catch `-*=*` which isn't matched by any of the above. (if the user accidentally misspells an option, they should get a reasonable error message).

configure
```diff
@@ -0,0 +1,363 @@
...
+fi
+
+## Resolve final paths
+SBINDIR="$EPREFIX/$SBINDIR"
+MANDIR="$PREFIX/$MANDIR"
+
+## Finalize $CXXFLAGS & $LDFLAGS
+if [ -z "${CXXFLAGS+IS_SET}" ]; then
+    CXXFLAGS=""
+    NO_CXXFLAGS="false"
```
davmac314, line 248: NO_CXXFLAGS gets set `false` when CXXFLAGS is unset - shouldn't it be `true` in that case? (or rename it `CXXFLAGS_SET` for example).

Makefile
```diff
@@ -28,14 +28,13 @@ clean:
 	$(MAKE) -C doc/manpages clean
 
 mconfig:
-	@UNAME=`uname`;\
+	@UNAME="INVALID";\
```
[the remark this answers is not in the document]
mobin-2008: This reserved for your last review and it's will be removed when no comments left.

configure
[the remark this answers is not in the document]
eli-schwartz: Keep in mind https://mywiki.wooledge.org/BashFAQ/105

`set -e` often doesn't do what people expect.

configure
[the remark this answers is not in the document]
eli-schwartz: autotools specifically describes --target as only useful or meaningful when compiling a compiler.

You build on the --build OS, you cross compile to produce binaries that run on the --host OS, and when you copy those built binaries over to the host OS and run them there, they then output machine code for the --target OS.

In nearly all cases you only care about the build and host OSes.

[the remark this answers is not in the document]
eli-schwartz: BTW: $HOSTCXX is for the --build machine, and autotools would call it $CXX_FOR_BUILD.

configure
```diff
@@ -0,0 +1,363 @@
...
+
+create_testfile_cc()
+{
+    echo "int main(int argc, char **argv) { return 0; }" > testfile.cc
+}
+
+findcxx()
+{
+    if type "$1" > /dev/null 2>&1; then
+        if [ "${2:-}" = "host" ]; then
```
davmac314, line 65: This function is never used with `host` as an argument, I think you can remove this check

configure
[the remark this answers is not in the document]
davmac314: Yeah this is true, autotools has very specific host/build/target meaning that we are conflicting with here. (That's my bad, I think of it in terms of "build native code for the machine that _hosts_ the build", but that's not the autotools usage). It would be better not to re-use the same names as autotools but in a different way; that could cause confusion.

I don't like "CXX_FOR_BUILD" much since it's ambiguous. But it's hard to come up with a short name that isn't. So I guess we may as well go with this.

@mobin-2008 let's rename HOSTCXX etc to CXX_FOR_BUILD - I'll do the same in existing mconfig files / makefiles, if you make the same changes here. So it will be:

CXX_FOR_BUILD
CXXFLAGS_FOR_BUILD
CPPFLAGS_FOR_BUILD
LDFLAGS_FOR_BUILD

I notice the help output only has HOSTCXX ( = CXX_FOR_BUILD), you should probably add the others also.

configure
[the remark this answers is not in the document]
davmac314: I'm aware of the issues with `set -e`, but using it still gives better behaviour than not using it in my opinion.

configure
mobin-2008: Seems like `NO_CXXFLAGS` is not a good name. I rename it to `AUTO_CXXFLAGS`.

configure
```diff
@@ -0,0 +1,361 @@
+#!/bin/sh
+### Custom configure script of Dinit.
+## Initial prepartion
+set -eu
+cd "$(dirname "$0")"
+
+## Helper functions
+# According to POSIX, echo has unspecified behavior if any of its arguments contain “\” or
```
davmac314, line 8: This text is copied verbatim from the linked page - please don't do that, it raises copyright issues (unless you have specific permission to use the text). Just summarise the issue in your own words; a short sentence or two is fine.

davmac314:
I just pushed 4ea7d5ddaa7a83662b01a12336c513f468b717ef to rename `HOSTxxx` to `xxx_FOR_BUILD` as discussed above, sorry for creating more work for you.

mobin-2008:
I fix many things reported by Davin and Eli. Also I fix some another things. If no comments left and CI report "success" status, I will remove "INVALID" in Makefile and merge it.

@davmac314 & @eli-schwartz, Thanks for code review :)

configure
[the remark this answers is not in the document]
mobin-2008: From [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/):
> -e
    When this option is on, when any command fails (for any of the reasons listed in [Consequences of Shell Errors](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_08_01) or by returning an [exit](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#exit) status greater than zero), the shell immediately shall exit, as if by executing the exit special built-in utility with no arguments, with the following exceptions:
>  * The failure of any individual command in a multi-command pipeline shall not cause the shell to exit. Only the failure of the pipeline itself shall be considered.
>  * The -e setting shall be ignored when executing the compound list following the while, until, if, or elif reserved word, a pipeline beginning with the ! reserved word, or any command of an AND-OR list other than the last.
> * If the exit status of a compound command other than a subshell command was the result of a failure while -e was being ignored, then -e shall not apply to this command.
> * This requirement applies to the shell environment and each subshell environment separately. For example, in:
> ` set -e; (false; echo one) | cat; echo two`
> the [false](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/false.html) command causes the subshell to exit without executing echo one; however, echo two is executed because the exit status of the pipeline (false; echo one) | cat is zero.

According to this, I prefer to use `set -e`.

configure
```diff
@@ -0,0 +1,372 @@
...
+        --disable-sanitize|--enable-sanitize=no) NO_SANITIZE=1 ;;
+        --enable-utmpx|--enable-utmpx=yes) USE_UTMPX=1 ;;
+        --disable-utmpx|--enable-utmpx=no) USE_UTMPX=0 ;;
+        CXX=*|CXX_FOR_BUILD=*|CXXFLAGS_FOR_BUILD=*|CPPFLAGS_FOR_BUILD=*\
+        |LDFLAGS_FOR_BUILD=*|CXXFLAGS=*|CXXFLAGS_EXTRA=*|TEST_CXXFLAGS=*\
+        |TEST_CXXFLAGS_EXTRA|LDFLAGS=*|LDFLAGS_EXTRA=*|TEST_LDFLAGS=*\
+        |TEST_LDFLAGS_EXTRA=*|CPPFLAGS=*)
+            envvar=$(expr "x$arg" : 'x\([^=]*\)=')
+            envvar_value=$(expr "X$arg" : '[^=]*=\(.*\)') || envvar_value="" # Fix empty var bug
+            eval "$envvar=$envvar_value"
```
davmac314, line 208: This was correct before. Now (without the backslash) it is not correct. Try it with `./configure CXXFLAGS="-Os -Wall"` and you'll see.

configure
```diff
@@ -0,0 +1,372 @@
...
+    fi
+    if [ "$PLATFORM" != "Darwin" ]; then
+        try_cxx_argument CXXFLAGS -fno-rtti
+    fi
+fi
+if [ "$PLATFORM" = "FreeBSD" ] && [ "$AUTO_LDFLAGS" = "true" ]; then
+    try_ld_argument LDFLAGS -lrt
+fi
+if [ "${NO_SANITIZE:-}" = "auto" ] && [ "$AUTO_TEST_LDFLAGS" = "true" ]; then
+    try_ld_argument TEST_LDFLAGS -fsanitize=address,undefined
```
davmac314, line 313: `-fsanitize` options are needed for compilation as well, not just linking

mobin-2008: I think if a flag can pass `try_ld_argument()`, It can added to `$CXXFLAGS` without test it with `try_cxx_argument()`. Right?

davmac314: Yes, `try_ld_argument` is doing compilation and linking so any argument that works is usable for both.

(This doesn't extend to all arguments though. For instance we don't want to put `-lrt` on the compiler command line even if it can be used when linking.)

configure
```diff
@@ -0,0 +1,385 @@
...
+        try_cxx_argument CXXFLAGS -fno-rtti
+    fi
+fi
+if [ "$PLATFORM" = "FreeBSD" ] && [ "$AUTO_LDFLAGS" = "true" ]; then
+    try_ld_argument LDFLAGS -lrt
+fi
+if [ "${NO_SANITIZE:-}" = "auto" ] && [ "$AUTO_TEST_LDFLAGS" = "true" ] && \
+   [ "$AUTO_TEST_CXXFLAGS" = "true" ]; then
+    try_ld_argument TEST_LDFLAGS -fsanitize=address,undefined
+    if [ "$AUTO_TEST_LDFLAGS" = "-fsanitize=address,undefined" ]; then
```
davmac314, line 321: I don't see how this fixes it. Am I missing something? (Did you test this change)?

configure
```diff
@@ -0,0 +1,385 @@
...
+        --enable-utmpx|--enable-utmpx=yes) USE_UTMPX=1 ;;
+        --disable-utmpx|--enable-utmpx=no) USE_UTMPX=0 ;;
+        CXX=*|CXX_FOR_BUILD=*|CXXFLAGS_FOR_BUILD=*|CPPFLAGS_FOR_BUILD=*\
+        |LDFLAGS_FOR_BUILD=*|CXXFLAGS=*|CXXFLAGS_EXTRA=*|TEST_CXXFLAGS=*\
+        |TEST_CXXFLAGS_EXTRA|LDFLAGS=*|LDFLAGS_EXTRA=*|TEST_LDFLAGS=*\
+        |TEST_LDFLAGS_EXTRA=*|CPPFLAGS=*)
+            envvar=$(expr "x$arg" : 'x\([^=]*\)=')
+            envvar_value=$(expr "X$arg" : '[^=]*=\(.*\)') || envvar_value="" # Fix empty var bug
+            eval "$envvar=\$envvar_value"
+            export "${envvar?}"
```
davmac314, line 209: Now you have restored the `export` as well. Why? It's not needed.

mobin-2008: When I just use `eval`, Users can't pass multi flags as `CXXFLAGS` & ....
I think this `export` is a simple way to fix this thing.

configure
mobin-2008: `$AUTO_TEST_LDFLAGS` is incorrect. It's must be `$TEST_LDFLAGS`. fixed

configure
davmac314: Can you give a full example that doesn't work without the `export`? What I've tried works for me without `export`.

(`export` exports the environment variables to subprocesses, I can't see why that would make any difference here.).

Before, when you removed `export`, you also removed the `\` in the `eval` line, which is why it stopped working then. Are you getting mixed up due to that perhaps? You need to remove `export` but keep the `eval` line exactly as it is.

mobin-2008: I think that's `\` for be processed by `export` but seems like I did wrong.

configure
```diff
@@ -0,0 +1,384 @@
...
+            envvar_value=$(expr "X$arg" : '[^=]*=\(.*\)') || envvar_value="" # Fix empty var bug
+            eval "$envvar=\$envvar_value"
+        ;;
+        *=*) warning Unknown variable: "$(expr "x$arg" : 'x\([^=]*\)=')" ;;
+        *) warning Unknown argument: "$arg" ;;
+    esac
+done
+
+## General Defines
+: "${PLATFORM:=$(uname)}"
```
mobin-2008, line 216: Now `configure` supports all variables be imported from environment. It's a good behavior? Can this be a problem?

configure
```diff
@@ -0,0 +1,384 @@
...
+        --disable-cgroups|--enable-cgroups=no) SUPPORT_CGROUPS=0 ;;
+        --enable-sanitize|--enable-sanitize=yes) NO_SANITIZE=0 ;;
+        --disable-sanitize|--enable-sanitize=no) NO_SANITIZE=1 ;;
+        --enable-utmpx|--enable-utmpx=yes) USE_UTMPX=1 ;;
+        --disable-utmpx|--enable-utmpx=no) USE_UTMPX=0 ;;
+        CXX=*|CXX_FOR_BUILD=*|CXXFLAGS_FOR_BUILD=*|CPPFLAGS_FOR_BUILD=*\
+        |LDFLAGS_FOR_BUILD=*|CXXFLAGS=*|CXXFLAGS_EXTRA=*|TEST_CXXFLAGS=*\
+        |TEST_CXXFLAGS_EXTRA|LDFLAGS=*|LDFLAGS_EXTRA=*|TEST_LDFLAGS=*\
+        |TEST_LDFLAGS_EXTRA=*|CPPFLAGS=*)
+            envvar=$(expr "x$arg" : 'x\([^=]*\)=')
```
davmac314, line 206: Why do you need expr here? Just use shell parameter expansion with suffix removal. Something like `envvar=${arg%%=*}`

configure
davmac314: I think it's ok as long as all environment variables are documented (eg in `--help` output)

davmac314: So it should be documented that PLATFORM, PREFIX, EPREFIX etc, take their values from the environment (and what this means). Or otherwise, don't take these values from the environment!

mobin-2008: So I think it's better to don't take values from environment.

configure
```diff
@@ -0,0 +1,375 @@
...
+if [ "$AUTO_CXXFLAGS" = "true" ]; then
+    for argument in -std=c++11 \
+                    -Wall \
+                    -Os \
+                    -fno-plt
+    do
+        try_cxx_argument CXXFLAGS $argument
+    done
+    if [ "$PLATFORM" != "FreeBSD" ]; then
+        try_cxx_argument CXXFLAGS -flto
```
davmac314, line 299: `-flto` needs to be passed to linker as well as compiler, and in that case, _all_ the optimisation options also need to be passed to the compiler.

configure
davmac314: Yep, ok

configure
davmac314: (As a consequence: you probably should only try `-flto` if both `AUTO_CXXFLAGS` and `AUTO_LDFLAGS` are true).

Also: there's no harm in trying to use LTO on FreeBSD, since if it fails you will not apply the flags anyway. That's the whole point of testing them :) (The `-fno-rtti` case is different because it won't cause compilation failure, but causes failure at runtime).

configure
```diff
@@ -0,0 +1,404 @@
...
+if [ "$AUTO_LDFLAGS" = "true" ]; then
+    if [ "$PLATFORM" = "FreeBSD" ]; then
+        try_ld_argument LDFLAGS -lrt
+    fi
+fi
+if [ "${NO_SANITIZE:-}" = "auto" ]; then
+    if [ "$AUTO_TEST_LDFLAGS" = "true" ] && [ "$AUTO_TEST_CXXFLAGS" = "true" ]; then
+        try_both_argument TEST_CXXFLAGS TEST_LDFLAGS -fsanitize=address,undefined
+    elif [ "$AUTO_TEST_LDFLAGS" = "true" ]; then
+        try_ld_argument TEST_LDFLAGS -fsanitize=address,undefined
```
davmac314, line 341: Does it really make sense to try this as a link-only argument if not adding it to compilation flags? You would be relying the sanitization flags being specified in the compiler flags by the user, in that case. Same below: does it make sense to add it as a compilation flag if it's not being added in the link flags?

More generally, there's an issue that `--enable-sanitize` doesn't work if TESTCXXFLAGS and TESTLDFLAGS are specified. This needs to be either documented properly or fixed (or, remove the option).

configure
```diff
@@ -0,0 +1,404 @@
...
+# General build options.
+
+# Linux (GCC): Note with GCC 5.x/6.x you must use the old ABI, with GCC 7.x you must use
+# the new ABI. See BUILD file for more information.
+# MacOS: Cannot use -fno-rtti: apparently prevents exception handling from working properly.
+# FreeBSD: Cannot use LTO with default linker.
+
+CXX=$CXX
+CXXFLAGS=$CXXFLAGS $CXXFLAGS_EXTRA
+TEST_CXXFLAGS=$CXXFLAGS $TEST_CXXFLAGS $TEST_CXXFLAGS_EXTRA
```
davmac314, line 371: So the `TEST_CXXFLAGS` used for build will always contain `CXXFLAGS` _as well as_ `TEST_CXXFLAGS` (and `TEST_CXXFLAGS_EXTRA`) passed into configure? Isn't the point of TEST_CXXFLAGS when passed into configure to be able to override the normal flags?

configure
```diff
@@ -0,0 +1,404 @@
...
+CPPFLAGS=$CPPFLAGS
+LDFLAGS=$LDFLAGS $LDFLAGS_EXTRA
+TEST_LDFLAGS=$LDFLAGS $TEST_LDFLAGS $TEST_LDFLAGS_EXTRA
```
davmac314, line 374: Same here - `TEST_LDFLAGS` is including `LDFLAGS`

configure
davmac314: So I've thought about this a little. Maybe you shouldn't accept `TEST_CXXFLAGS`/`TEST_LDFLAGS` as user input in the script? You can accept `TEST_CXXFLAGS_EXTRA`/`TEST_LDFLAGS_EXTRA` only, for example. That would make things a lot simpler.

Otherwise you should:

- fix: not include `CXXFLAGS` in the output `TEST_CXXFLAGS`, and
- fix: not include `LDFLAGS` in the output `TEST_LDFLAGS`, and
- ensure that `--enable-sanitize` does something sensible when `TEST_CXXFLAGS` and/or `TEST_LDFLAGS` are specified directly by the user. I.e. adds the flags anyway (with a message printed) or does nothing (with a message printed).

User experience is important! If something looks like it should work, then either it should work, or it should at least give feedback rather than silently not working.

Eg consider right now if I:

```
./configure TEST_CXXFLAGS="" TEST_LDFLAGS="" --enable-sanitizer
```

There's nothing in the help output that suggests this won't work - but it won't!

configure
mobin-2008: So just try `-fsanitize` when both of them are `auto`?

davmac314: See my larger note against line 371. Just trying `-fsanitize` when both `AUTO_TEST_LDFLAGS` and `AUTO_TEST_CXXFLAGS` are true is one option, but then `--enable-sanitize` should do something sensible (eg: either print an error message, or add `-fsanitize` unconditionally and print a message that `TEST_LDFLAGS`/`TEST_CXXFLAGS` were modified if they were specified by the user)

configure
mobin-2008: I removed `--enable-sanitize` option and `TEST_CXXFLAGS` doesn't include `CXXFLAGS`. but seems like it's break Unit tests on MacOS.

davmac314: The _default_ TEST_CXXFLAGS should be to use CXXFLAGS. You need something like:

```
if [ "$AUTO_TEST_LDFLAGS" = "true" ]; then
    TEST_LDFLAGS=$LDFLAGS
fi
if [ "$AUTO_TEST_CXXFLAGS" = "true" ]; then
    TEST_CXXFLAGS=$CXXFLAGS
fi
```

... between lines 330-331.

davmac314:
Looks good, if you remove the testing change (`@UNAME="INVALID"`) we can merge

davmac314 (approved):

Issue #109, opened 2022-09-12
improve service environment variables


Duncaen:
Here are a few features I would like to see regarding environment variables, and I would be open to contribute them, but would like to have some feedback before I would work on them.

## streamlined environment handling

I think it would be nice if variables from `env-file` in the command line and and when resolving `socket-listen`, `logfile`, `working-dir` and `pid-file`, instead of just the `dinit` processes environment.

Two alternative approaches for loading `env-file`
1. when the service is loaded
2. when the service starts, and then resolve paths and the command line.

I would prefer 1. as this would allow linting with `dinitcheck` and would open the door for a `dinitctl` command to inspect and set per service variables:

```
dinitctl setenv [--service <service name>] [name[=value] ...]
dinitctl getenv [--service <service name>] [name ...]
```

## default user variables

Additionally I think there are a few variables that should be initialized by default based on the `run-as` users `/etc/passwd` entry:
* `HOME`
* `LOGNAME`
* `SHELL`

I think those should be set before loading `env-file`, so that `env-file` could reference the right `HOME`.
The `env-file` could then unset or overwrite those default variables if required.

## basic substitutions

Another feature I think would be nice to have is basic posix shell variable substitutions in settings and in the `env-file`.

The ones I think are the most useful:
```
${VAR-value if var is unset or null}
${VAR+alternative value if var is set}
```

Checking that a variable is set might be useful, but is not super important:
```
${VAR?[error if var is not set with optional description]}
```

I think `${VAR%...}` and `${VAR#...}` are probably not useful enough to justify the extra requirement for a posix pattern matching implementation.

davmac314:
> Two alternative approaches for loading env-file
> 
> 1.     when the service is loaded
> 2.     when the service starts, and then resolve paths and the command line.
> 
> I would prefer 1. 
> 

Yes, I have been thinking about this for a while and would also prefer 1.

Functionality to set environment variables for specific services is also something I've been thinking about, so no objections there either.

> default user variables

I have no objection to any of these.

> basic substitutions

I am hesitant because the question soon becomes "where do you draw the line". However for the two that you highlight (`${VAR-...}` and `${VAR+...}`) I would accept a merge request.

Thanks for checking!

dkwo:
> I think those should be set before loading `env-file`, so that `env-file` could reference the right `HOME`.

Do I understand correctly that this feature is currently not present in dinit?
@davmac314 If so, are you open to someone implementing it?
I think this would help with user services.

Issue #53, opened 2022-04-11
[Feature Request] populate $SERVICE_NAME or logdir configuration


RichardKraus:
It would be helpful to make $SERVICE_NAME or similar expand to the service name on service load (if sub-vars is set).
This could eliminate the need to retype the logfile name for each service.

alternatively it could also be quite nice to have a logdir option (or handle dirs in logfiles) by creating a logfile `<service name>.log` in that directory.

If that is already a thing then I did not find it yet.

PR #144, opened 2023-01-10
substitute env-file in service files at load

q66:
The previous behavior did not substitute env-file variables at service load, because it could only see the environment available to dinit process. This changes that, and env-files are instead loaded during service load, and all variables are substituted.

The vars set up during process call are not exposed here. These would have limited usability in any case and would significantly complicate the whole thing.

I am putting this here to get comments. The code is not necessarily good/clean, e.g. dinitcheck is not properly implemented at this point, and there are no tests yet. I mostly want to know if continuing with this approach is okay. The referenced issue from @Duncaen also talks about implementing basic variable substitutions, I still gotta think about those. I think a well defined subset would be nice, though I see that being more useful in the service files themselves rather than in the env-file (e.g. to substitute defaults when there is no env-file or the variable is missing from env-file). I am also not implementing the default variables, I will do that once this is okay.

My original approach involved building the environment map once and then passing that to the child process, and doing the additional variable setting directly on that. However, this turned out not working because the desired environment may change between load and run through changes of base dinit process environment, e.g. via `dinitctl setenv`. If we were to do that we'd have to scan for all loaded service and update each's maps, which is no good.

Ref https://github.com/davmac314/dinit/issues/109

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 # Integration test output:
 /src/igr-tests/basic/basic-ran
 /src/igr-tests/environ/env-record
+/src/igr-tests/environ2/env-record
 /src/igr-tests/ps-environ/env-record
 /src/igr-tests/chain-to/recorded-output
 /src/igr-tests/restart/basic-ran
--- a/doc/manpages/dinit-service.5.m4
+++ b/doc/manpages/dinit-service.5.m4
@@ -160,14 +160,9 @@ group of the specified user.
 \fBenv\-file\fR = \fIfile\fR
 Specifies a file containing value assignments for environment variables, in the same
 format recognised by the \fBdinit\fR command's \fB\-\-env\-file\fR option (see \fBdinit\fR(5)).
-The file is read (or re-read) whenever the service is started; the values read do not
-affect for the processing performed for the \fBsub\-vars\fR load option, which is done
-when the service description is loaded.
-The precise behaviour of this setting may change in the future.
-It is recommended to avoid depending on the specified file contents being reloaded
-whenever the service process starts.
+The file is read when the service is loaded, therefore values from it can be used in substitutions.
 .sp
-The path specified is subject to variable substitution (see \fBVARIABLE SUBSTITUTION\fR).
+That means this path is static and cannot itself have variable substitutions (see \fBVARIABLE SUBSTITUTION\fR).
 .TP
 \fBrestart\fR = {yes | true | no | false}
 Indicates whether the service should automatically restart if it stops, including due to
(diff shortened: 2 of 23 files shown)


davmac314:
I think this is ok in principle. Having an environment-per-service was something I had already been thinking about (which is why you can layer environments already).

I haven't examined the code too closely since you made it clear that it's not finished, one thing I do note is that you may need to handle exceptions from `read_env_file` when you call it in `load_reload_service` and re-throw a better exception (doesn't really matter for bad_alloc, but for file read errors we want to report an error loading the environment file, not a general error reading the service description; I hope that makes sense).

q66:
btw, is there an *actual* reason for existence of the sub-vars option? imo it has some problems:

1) it only applies to commands, while everything else expands vars implicitly
2) there does not seem to be much purpose to it regardless; if you don't use variable expansion in the commands, the option just doesn't do anything; if you do want variable expansion to happen, you have to set it, but again I don't see much of a reason for having  to do so

davmac314:
> btw, is there an _actual_ reason for existence of the sub-vars option? imo it has some problems:

Yeah, it's a bit of historical baggage. When dinit itself didn't support variable substitution, you could get around that by invoking the shell via something likeL

`command = /bin/sh do-something $SOMEVAR`

when the substitution support was implemented, I didn't want to break existing service descriptions which did something like that, so I added `sub-vars` as an opt-in way to turn the substitution on (bear in mind you can do more advanced substitution with the shell, so even though the behaviour might anyway be the same for the above example, it wouldn't in any case. And, I didn't want to break any service description which had a command which happened to have a `$` for any other reason (though I agree that's probably rare).

We could get rid of that option now, I think.

q66:
wouldn't this not work anyway?

you would need something more like

```
command = sh -c "do-something $SOMEVAR"
```

because -c takes a single string, and in those cases dinit would *not* autosubstitute

davmac314:
Yes, it would need the quotes, and that would inhibit expansion by dinit 0.12 and prior, that's true. I'm having to check the history here. I didn't _remove_ the option in 0.13 because it would potentially break this type of command (which invokes the shell). The _original_ reason for `sub-vars` was to not break any service description written prior to adding support for the original substitution (0.2.0 apparently) and which happened to use an argument like `$something`. This is probably rare and worrying about this was probably excessively cautious. As I said already, i think it would be ok to get rid of the option.

q66:
great; I'm thinking of just implementing a common path for variable substitution for both commands and paths (just the syntax for $stuff, so that we do not have to duplicate the rules implementing that), and imo in those hypothetical cases when somebody wants an argument that is explicitly $foo they could just escape it (as is already supported in paths)

mobin-2008:
I think it's better to take a look to #53

q66:
i finally found some time to get back to this, so i'm working on completing it

q66:
I think this is mostly ready to get some initial review.

Functionality implemented:

* `sub-vars` is gone
* limited POSIX parameter expansion in service files
* default environment variables for `run-as` services (loaded before env file)
* env files are loaded at service load and one can use variables from them in services
* I have restricted env var names not to allow dashes or dots; they are normally not allowed and either break parsing or result in strange unexpected semantics

Still to be done:

* more tests maybe? i haven't tested the default run-as envvars code

q66:
i wonder if setting the user vars would be a good place for a `load-options` setting (with setting: always set, without setting: never set), and perhaps also expand it with `UID` and `GID` (as these are useful for path expansions)

davmac314:
> i wonder if setting the user vars would be a good place for a `load-options` setting (with setting: always set, without setting: never set), and perhaps also expand it with `UID` and `GID` (as these are useful for path expansions)

Yeah, that makes sense to me.

I'll look over the current changes as soon as I can.

q66:
Updated. I used `set-passwd-vars` as the option name, feel free to come up with something better (I'm not really sure).

q66:
I also tried addressing https://github.com/davmac314/dinit/issues/53 as a separate commit on top of this. If you don't like that one, I can discard it, but I think it could be moderately useful. In any case, I left it separate (but it builds on the other functionality). It is activated with a load option too, `export-service-name`.

davmac314 (approved):

PR #171, opened 2023-04-02
dinitctl: add is-active and is-failed commands

q66:
The is-active command makes dinitctl return 0 when the given service is currently active, and 1 in any other scenario. The is-failed command makes it return 0 when the given service has failed to start and is currently stopped, and 1 in any other scenario. The service not existing and so on counts as non-success for both commands.

The purpose of these is to use them in scripts, as parsing status output is clunky. In general it is enough to pass --quiet and just check the status code, but both commands also print the current service status (and nothing else) if in default verbose mode.

It is similar to the same-named commands in `systemctl`.

### DIFF
--- a/doc/manpages/dinitctl.8.m4
+++ b/doc/manpages/dinitctl.8.m4
@@ -18,6 +18,12 @@ dinitctl \- control services supervised by Dinit
 [\fIoptions\fR] \fBstatus\fR \fIservice-name\fR
 .HP
 .B dinitctl
+[\fIoptions\fR] \fBis-active\fR \fIservice-name\fR
+.HP
+.B dinitctl
+[\fIoptions\fR] \fBis-failed\fR \fIservice-name\fR
+.HP
+.B dinitctl
 [\fIoptions\fR] \fBrestart\fR [\fB\-\-no\-wait\fR] [\fB\-\-ignore\-unstarted\fR] \fIservice-name\fR
 .HP
 .B dinitctl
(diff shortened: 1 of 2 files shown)


mobin-2008:
Hello. These commands are unnecessary in my view. I suggest to just add some special return codes to `dinitctl status` command instead of creating some new commands called `is-active/failed`.

For example:
if service was loaded, started & is active return `0`.
if service was loaded, started & is inactive return `1`.
if service was loaded, but stuck in `STARTING`/`STOPPED` return `3`. (I'm not sure about this)
if service was not loaded return `4`.

but it's need to add `--quiet` support to `dinitctl status`.
Regards

q66:
they are not unnecessary at all

a special return code is 1) less intuitive 2) still clunkier to check 3) changes the existing behavior of status

a quiet mode for status will always mean uglier syntax, because 1) if you make quiet status just print out the status word, you will have to manually silence the output AND check the return code explicitly, so instead of `if dinitctl --quiet is-active foo; then ...` you will get at very least `dinitctl status foo > /dev/null; if [ $? -eq whatever ]; then ...` 2) if you make --quiet print nothing, you will lose a reasonable way to check only the current status, and you will still have to test for the return code separately

having a separate command not only makes for much nicer checks without having to document special return codes (which is inconsistent with behavior of every other dinitctl command, too) but also it makes for an unambiguous way to check this without conflicting with protocol errors and so on (and it provides a syntax people are already familiar with as systemctl does the same thing, and even outside of that, it's much easier to tell what it does because special return codes are not self-describing, while something like is-active is)

davmac314 (approved):

davmac314:
I agree, it's useful and a better user experience to be able to check specific conditions by name rather than status codes.

Issue #134, opened 2022-12-14
[feature request] sending signals to services


CMB:
It would be nice if I could use dinitctl to signal a service.
Like with runit, I can say `sv hup lighttpd` to tell lighttpd to reload its config.

mobin-2008:
Hi. I thinked about it. Most UNIX-like systems (such as Linux based and BSDs and even Solaris) have `kill` and `pkill`. `pkill` is very easy to use:
````
# for example tell nginx to reload its config
pkill -HUP nginx
````
I don't see a special case to need a special command to send a signal to services because `kill` and `pkill` are exist.

CMB:
Yes kill and pkill will do it, but they may cause collateral damage.
E.G., multiple nginx services running on one system.
The kill and pkill commands operate at the wrong granularity.

Better than pkill would be to ask dinit for the pid, like so:

    kill -HUP "$( dinitctl list |grep ' postgres ' |sed -E 's;^.*pid: ([0-9]+).*;\1;')"

Signaling processes is also racy when you aren't the parent.
There's a race window between the time you obtain the PID and the time you signal the process,
during which the PID might have been recycled.  Granted, the window is pretty tiny, but the kill command is still an act of faith.
With a supervised service, `type=process` in dinit, dinit is the parent, and it can signal the appropriate child in a race-free way.

davmac314:
Note that the usual policy is not to accept feature requests without a patch (see the README) but in this case, yes, it's something that is planned.

james-knippes:
I've implemented signal support in #176 in hopes it may be useful. I would appreciate feedback. I assume the implementation covers your request @CMB ?

mobin-2008:
I forgot to close this issue. This feature has been added in #176 and will be available in 0.17 and later.

PR #176, opened 2023-04-19
Feature signal support

james-knippes:
This adds support for sending signals to services, as I consider it a generally useful feature and implemented it with hacky scripts parsing the output of `dinitctl status $service` in the past. Many services use signals for reloading configuration files and/or enabling/disabling specific runtime features. The feature is currently requested by #134 and @davmac314 commented that
> it's something that is planned

As this would be my first contribution to the project I appreciate any feedback. I'm willing to implement changes/improvements.

I've tried to follow contribution guidelines. Tests are still missing. I will add them if the feature is generally considered accepted.
 - [x]  should follow existing style and conventions
 - [x] should keep the design considerations in mind (read the DESIGN document)
 - [x] should be accompanied by appropriate documentation additions/updates
 - [ ] should include tests where practical
 
Open questions/limitations:
- Should signalling be moved from `control.cc` directly as a feature into the (decedents of the) `service_record` class ?
- I've found no way to translate signal names/abbrevations to their integer counterparts at runtime, that is system and architecture independent. The [`sigabbrev_np`](https://man7.org/linux/man-pages/man3/strsignal.3.html) function is a gnu extension and only works for non realtime signals. [Here](https://github.com/util-linux/util-linux/blob/master/lib/signames.c#L130) is how `kill` from `util-linux` does it, but I'm not sure if this kind of conversion is worth it.

### DIFF
--- a/doc/manpages/dinitctl.8.m4
+++ b/doc/manpages/dinitctl.8.m4
@@ -70,6 +70,9 @@ dinitctl \- control services supervised by Dinit
 .HP
 .B dinitctl
 [\fIoptions\fR] \fBcatlog\fR [\fB--clear\fR] \fIservice-name\fR
+.HP
+.B dinitctl
+[\fIoptions\fR] \fBsignal\fR ( \fB\-\-list/\-l\fR | \fIsignal-number\fR \fIservice-name\fR )
 .\"
 .PD
 .hy
@@ -313,6 +316,18 @@ Show the contents of the log buffer for the named service.
 This is possible only if the log type of the service is set to \fBbuffer\fR.
 If the log is truncated or appears incomplete, a warning message follows the output.
 If the \fB\-\-clear\fR option is specified, the buffer is cleared after displaying its contents. 
+.TP
+\fBsignal\fR
+Send a signal to a specified service.
+The \fIsignal-number\fR can either be specified as integer or signal name.
+Signal names are only defined for these signals:
+\fBHUP\fR, \fBKILL\fR, \fBUSR1\fR, \fBUSR2\fR, \fBSTOP\fR, \fBINT\fR, \fBTERM\fR,
+\fBCONT\fR, \fBQUIT\fR and if system supports it; \fBINFO\fR.
+Other system specific signals, like Linux real-time signals, have to be specified as integer.
+Dinit will accept any signal which supported by \fBkill(2)\fR syscall.
+.sp
+The \fB--list\fR or \fB-l\fR option can be used (instead of \fIsignal-number\fR and \fIservice-name\fR)
+to list available signal names and their corresponding integer value for the current platform/system.
 .\"
 .SH SERVICE OPERATION
 .\"
--- a/src/control.cc
+++ b/src/control.cc
@@ -12,7 +12,7 @@
 
 // Control protocol versions:
 // 1 - dinit 0.16 and prior
-// 2 - dinit 0.17 (adds DINIT_CP_SETTRIGGER, DINIT_CP_CATLOG)
+// 2 - dinit 0.17 (adds DINIT_CP_SETTRIGGER, DINIT_CP_CATLOG, DINIT_CP_SIGNAL)
 
 namespace {
     constexpr auto OUT_EVENTS = dasynq::OUT_EVENTS;
@@ -118,6 +118,9 @@ bool control_conn_t::process_packet()
     }
     if (pktType == DINIT_CP_CATLOG) {
         return process_catlog();
+    }
+    if (pktType == DINIT_CP_SIGNAL) {
+        return process_signal();
     }
 
     // Unrecognized: give error response
@@ -985,6 +988,54 @@ bool control_conn_t::process_catlog()
         bps->clear_log_buffer();
     }
     return queue_packet(std::move(pkt));
+}
+
+bool control_conn_t::process_signal()
+{
+    // packet contains signal number and process handle
+    constexpr int pkt_size = 1 + sizeof(int) + sizeof(handle_t);
+    if (rbuf.get_length() < pkt_size) {
+        chklen = pkt_size;
+        return true;
+    }
+
+    int sig_num;
+    rbuf.extract((char *) &sig_num, 1, sizeof(sig_num));
+    handle_t handle;
+    rbuf.extract((char *) &handle, 1 + sizeof(sig_num), sizeof(handle));
+    rbuf.consume(pkt_size);
+    chklen = 0;
+
+    service_record *service = find_service_for_key(handle);
+    if (service == nullptr) {
+        char nak_rep[] = { DINIT_RP_NAK };
+        return queue_packet(nak_rep, 1);
+    }
+
+    // Reply:
+    // 1 byte packet type = DINIT_RP_*
+
+    pid_t spid = service->get_pid();
+    // we probably don't want to kill/signal every process (in the current group),
+    // but get_pid() sometimes returns -1 if e.g. service is not 'started'
+    if (spid == -1 || spid == 0) {
+        char nak_rep[] = { DINIT_RP_SIGNAL_NOPID };
+        return queue_packet(nak_rep, 1);
+    }
+    else {
+        if (bp_sys::kill(spid, sig_num) != 0) {
+            if (errno == EINVAL) {
+                log(loglevel_t::ERROR, "Requested signal not in valid signal range.");
+                char nak_rep[] = { DINIT_RP_SIGNAL_BADSIG };
+                return queue_packet(nak_rep, 1);
+            }
+            log(loglevel_t::ERROR, "Error sending signal to process: ", strerror(errno));
+            char nak_rep[] = { DINIT_RP_SIGNAL_KILLERR };
+            return queue_packet(nak_rep, 1);
+        }
+    }
+    char ack_rep[] = { DINIT_RP_ACK };
+    return queue_packet(ack_rep, 1);
 }
 
 bool control_conn_t::query_load_mech()
--- a/src/dinitctl.cc
+++ b/src/dinitctl.cc
@@ -53,6 +53,8 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, const char *
 static int do_setenv(int socknum, cpbuffer_t &rbuffer, std::vector<const char *> &env_names);
 static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service_name, bool trigger_value);
 static int cat_service_log(int socknum, cpbuffer_t &rbuffer, const char *service_name, bool do_clear);
+static int signal_send(int socknum, cpbuffer_t &rbuffer, const char *service_name, int sig_num);
+static int signal_list();
 
 static const char * describeState(bool stopped)
 {
@@ -85,6 +87,8 @@ enum class command_t {
     SET_TRIGGER,
     UNSET_TRIGGER,
     CAT_LOG,
+    SIG_SEND,
+    SIG_LIST,
     IS_STARTED,
     IS_FAILED,
 };
@@ -117,6 +121,9 @@ int dinitctl_main(int argc, char **argv)
     bool do_force = false;
     bool ignore_unstarted = false;
     bool use_passed_cfd = false;
+    bool show_siglist = false;
+    std::string sigstr;
+    int sig_num = -1;
     
     command_t command = command_t::NONE;
 
@@ -193,6 +200,15 @@ int dinitctl_main(int argc, char **argv)
                     break;
                 }
             }
+            else if (strcmp(argv[i], "--list") == 0 || strcmp(argv[i], "-l") == 0) {
+                if (command == command_t::SIG_SEND) {
+                    show_siglist = true;
+                }
+                else {
+                    cmdline_error = true;
+                    break;
+                }
+            }
             else {
                 cerr << "dinitctl: unrecognized/invalid option: " << argv[i] << " (use --help for help)\n";
                 return 1;
@@ -262,6 +278,9 @@ int dinitctl_main(int argc, char **argv)
             else if (strcmp(argv[i], "catlog") == 0) {
                 command = command_t::CAT_LOG;
             }
+            else if (strcmp(argv[i], "signal") == 0) {
+                command = command_t::SIG_SEND;
+            }
             else {
                 cerr << "dinitctl: unrecognized command: " << argv[i] << " (use --help for help)\n";
                 return 1;
@@ -304,6 +323,22 @@ int dinitctl_main(int argc, char **argv)
                 }
                 to_service_name = argv[i];
             }
+            else if (command == command_t::SIG_SEND) {
+                if (!show_siglist) {
+                    if (sigstr.empty()) {
+                        sigstr = argv[i];
+                    }
+                    else if (service_name == nullptr) {
+                        service_name = argv[i];
+                    }
+                    else {
+                        cmdline_error = true;
+                    }
+                }
+                else {
+                    cmdline_error = true;
+                }
+            }
             else {
                 cmd_args.push_back(argv[i]);
             }
@@ -324,8 +359,50 @@ int dinitctl_main(int argc, char **argv)
             cmdline_error = true;
         }
     }
+    else if (command == command_t::SIG_SEND) {
+        if (show_siglist) {
+            if (sigstr.empty()) {
+                command = command_t::SIG_LIST;
+            }
+            else {
+                cmdline_error = true;
+            }
+        }
+        else {
+            if (sigstr.empty()) {
+                cerr << "dinitctl: signal number/name must be specified" << std::endl;
+                return 1;
+            }
+            if (service_name == nullptr) {
+                cerr << "dinitctl: service name must be specified" << std::endl;
+                return 1;
+            }
+            sig_num = dinit_load::signal_name_to_number(sigstr);
+            if (sig_num == 0) {
+                cerr << "dinitctl: '" << sigstr
+                        << "' is not a valid signal name/number" << std::endl;
+                return 1;
+            }
+            else if (sig_num == -1) {
+                try {
+                    size_t pos;
+                    sig_num = std::stoi(sigstr, &pos);
+                    if (sigstr.size() != pos) {
+                        throw std::invalid_argument("");
+                    }
+                } // catch both invalid_argument and out_of_range
+                catch (std::exception) {
+                    cerr << "dinitctl: '" << sigstr
+                            << "' is not a valid signal name/number" << std::endl;
+                    return 1;
+                }
+            }
+        }
+    }
     else {
-        bool no_service_cmd = (command == command_t::LIST_SERVICES || command == command_t::SHUTDOWN);
+        bool no_service_cmd = (command == command_t::LIST_SERVICES
+                              || command == command_t::SHUTDOWN
+                              || command == command_t::SIG_LIST);
         if (no_service_cmd) {
             if (!cmd_args.empty()) {
                 cmdline_error = true;
@@ -375,6 +452,7 @@ int dinitctl_main(int argc, char **argv)
           "    dinitctl [options] untrigger <service-name>\n"
           "    dinitctl [options] setenv [name[=value] ...]\n"
           "    dinitctl [options] catlog <service-name>\n"
+          "    dinitctl [options] signal <signal> <service-name>\n"
           "\n"
           "Note: An activated service continues running when its dependents stop.\n"
           "\n"
@@ -401,6 +479,11 @@ int dinitctl_main(int argc, char **argv)
         return 1;
     }
     
+    // SIG_LIST doesn't need a control socket connection so handle it specially.
+    if (command == command_t::SIG_LIST) {
+        return signal_list();
+    }
+
     // Begin the real work: connect to dinit
 
     signal(SIGPIPE, SIG_IGN);
@@ -483,6 +566,9 @@ int dinitctl_main(int argc, char **argv)
                 throw cp_old_server_exception();
             }
             return cat_service_log(socknum, rbuffer, service_name, catlog_clear);
+        }
+        else if (command == command_t::SIG_SEND) {
+            return signal_send(socknum, rbuffer, service_name, sig_num);
         }
         else {
             return start_stop_service(socknum, rbuffer, service_name, command, do_pin, do_force,
@@ -1810,6 +1896,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
         rbuffer.consume(1);
     }
 
+    return 0;
+}
+
+static int signal_send(int socknum, cpbuffer_t &rbuffer, const char *service_name, int sig_num)
+{
+    using namespace std;
+
+    handle_t handle;
+
+    if (!load_service(socknum, rbuffer, service_name, &handle, nullptr, true)) {
+        return 1;
+    }
+
+    // Issue SIGNAL command.
+    auto m = membuf()
+            .append<char>(DINIT_CP_SIGNAL)
+            .append(sig_num)
+            .append(handle);
+    write_all_x(socknum, m);
+
+    wait_for_reply(rbuffer, socknum);
+    if (rbuffer[0] == DINIT_RP_NAK) {
+        cerr << "dinitctl: cannot send signal to service.\n";
+        return 1;
+    }
+    if (rbuffer[0] == DINIT_RP_SIGNAL_NOPID) {
+        cerr << "dinitctl: could not get vaild PID of service; service is not process, "
+        "service in wrong state." << endl;
+        return 1;
+    }
+    if (rbuffer[0] == DINIT_RP_SIGNAL_BADSIG) {
+        cerr << "dinitctl: provided signal was invalid.\n";
+        return 1;
+    }
+    if (rbuffer[0] == DINIT_RP_SIGNAL_KILLERR) {
+        cerr << "dinitctl: failed sending signal to service.\n";
+        return 1;
+    }
+    if (rbuffer[0] != DINIT_RP_ACK) {
+        cerr << "dinitctl: protocol error.\n";
+        return 1;
+    }
+    rbuffer.consume(1);
+    return 0;
+}
+
+static int signal_list()
+{
+    using namespace std;
+    using namespace dinit_load;
+    cout << "dinitctl: The following signal names are supported:";
+    int skip_none = 0;
+    for (const auto &signal: signal_to_int_map) {
+        if (skip_none < 2) {
+            skip_none += 1;
+        }
+        else {
+            cout << endl << "dinitctl: ";
+            string sigpad = signal.first;
+            sigpad.resize(5,' ');
+            cout << sigpad << "-> " << signal.second;
+        }
+    }
+    cout << endl;
     return 0;
 }
 
--- a/src/includes/load-service.h
+++ b/src/includes/load-service.h
@@ -150,18 +150,36 @@ inline string_iterator skipws(string_iterator i, string_iterator end) noexcept
     return i;
 }
 
-// Convert a signal name to the corresponding signal number
-inline int signal_name_to_number(std::string &signame) noexcept
+// The list to converting some strings/signals to int
+// The signal list is created statically with only those signals,
+// which are defined for the current architecture/system
+const std::pair<const char *, int> signal_to_int_map[] = {
+    { "none", 0 },
+    { "NONE", 0 },
+    { "HUP", SIGHUP },
+    { "INT", SIGINT },
+    { "QUIT", SIGQUIT },
+    { "KILL", SIGKILL },
+    { "USR1", SIGUSR1 },
+    { "USR2", SIGUSR2 },
+    { "TERM", SIGTERM },
+    { "CONT", SIGCONT },
+    { "STOP", SIGSTOP },
+#ifdef SIGINFO
+    { "INFO", SIGINFO },
+#endif
+};
+
+inline int signal_name_to_number(const std::string &signame) noexcept
 {
-    if (signame == "none" || signame == "NONE") return 0;
-    if (signame == "HUP") return SIGHUP;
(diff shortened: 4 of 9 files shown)


mobin-2008:
Hello. Thanks for the PR. I did really quick review on it and I suggest to add `DINIT_CP_SIGNAL` as "new feature in Control protocol V2 (Dinit 0.17)" in line 15 of [`control.cc`](https://github.com/james-knippes/dinit/blob/7bc18fb698acc3f997d011edb19ffa713ef7ee45/src/control.cc#L15) file.

Regards

q66:
i think it would also be nice if there was a unified way to tell a service to reload its configuration, without sending an explicit signal; the service file could define the action (with choice between signal and command, e.g. `reload-signal = hup` or `reload-command = something`, default implicit behavior being sending SIGHUP)

unfortunately there is already `dinitctl reload`, just with a different meaning than what it means in most service managers...

james-knippes:
> Hello. Thanks for the PR. I did really quick review on it and I suggest to add `DINIT_CP_SIGNAL` as "new feature in Control protocol V2 (Dinit 0.17)" in line 15 of [`control.cc`](https://github.com/james-knippes/dinit/blob/7bc18fb698acc3f997d011edb19ffa713ef7ee45/src/control.cc#L15) file.

Thanks . I've added DINIT_CP_SIGNAL to the list of new features.

james-knippes:
> i think it would also be nice if there was a unified way to tell a service to reload its configuration, without sending an explicit signal; the service file could define the action (with choice between signal and command, e.g. `reload-signal = hup` or `reload-command = something`, default implicit behavior being sending SIGHUP)

I also think this would be a very useful feature. And one that a lot of service managers and services support. Signalling services may have a lot different uses though. I use it e.g. for toggling the visibility of my [onscreen keyboard](https://git.sr.ht/~proycon/wvkbd) with SIGRTMIN+0.

The (implementation details of a) service reloading feature probably deserves its own discussion, as it may impact future usability/expendability. I've started #178 for this purpose (:

james-knippes:
Could someone point me to the right place and to which extend tests should be implemented for this feature?
I'm not too familiar with (c++) testing in general. I would guess in test.cc or proctests.cc, but i'm not sure how to test the handling of pid values, as they don't seem to get mocked in the `test_service` or `base_process_service_test` class ?

I may be missing some basics here and would appreciate hints for test examples for orientation or general resources/introductions to the topic.

davmac314:
Sorry for taking a little while to get to looking at your PR properly, and thanks for offering a contribution!

The actual code changes look good (minus a few minor style inconsistencies eg spaces around the condition in `if`).

My main concern is that being forced to specify signals via number rather than name is awkward. There is already a `signal_name_to_number` function in load-service.h, which is already included by dinitctl.cc so would be straightforward to use.

Also the default of USR1 (in fact any default) is unlikely to be useful for most services, so it would be a better interface if the signal could just be specified without a switch, something like:

```
dinitctl signal <signal-name-or-number> <service-name>
```

You could potentially still allow just:

```
dinitctl signal <service-name>
```

... to send a default signal (USR1 or whatever) but in my opinion this isn't really necessary.

davmac314:
> Could someone point me to the right place and to which extend tests should be implemented for this feature?
> I'm not too familiar with (c++) testing in general. I would guess in test.cc or proctests.cc, but i'm not sure how to test the handling of pid values, as they don't seem to get mocked in the `test_service` or `base_process_service_test` class ?

Probably proctests.cc.

Pids are mocked via test-bpsys.cc which also contains mocks for various system functions like `kill`. Within proctests.cc you can use `last_sig_sent` to check what signal has most recently been sent (there's currently no way to check whether it was sent to the correct pid, so you could not worry about that for now, or make changes in test-bpsys.cc so that it records the pid target in another variable, or implement an integration test instead).

davmac314:
> Probably proctests.cc.

Actually, since the only thing to really test is the control protocol handling, probably cptests.cc. The complication is that this test needs to use a process-based service which the cptests currently don't do. The proctests should give you an idea of what's necessary; you have to call `base_process_service_test::exec_succeeded` as part of the test to actually get the test service started. See `test_proc_service_start()` in proctests.cc for that.

james-knippes:
@davmac314 thank you for the feedback.
> The actual code changes look good (minus a few minor style inconsistencies eg spaces around the condition in `if`).

I've fixed the white-space inconsistencies.

> Also the default of USR1 (in fact any default) is unlikely to be useful for most services, so it would be a better interface if the signal could just be specified without a switch, something like:
> ```
> dinitctl signal <signal-name-or-number> <service-name>
> ```

You are totally right, the interface was rather awkward. I've update the the parameter handling accordingly

> My main concern is that being forced to specify signals via number rather than name is awkward. There is already a `signal_name_to_number` function in load-service.h, which is already included by dinitctl.cc so would be straightforward to use.

I don't know how i could miss the existing `signal_name_to_number` function :D I've updated dinitctl to accept the signal names or numbers/integers. I've also changed the `signal_name_to_number` function to include all POSIX signals. System specific signals like Linux real-times signals have to be entered as integers.

I'll have a look at the tests in the next days.

davmac314:
Hi @james-knippes - just wondering if you were still planning on finishing off this PR? I think that tests are the only thing pending

src/tests/cptests/cptests.cc
```diff
@@ -951,6 +1012,52 @@ void cptest_servicestatus()
...
+
+    int fd = bp_sys::allocfd();
+    auto *cc = new control_conn_t(event_loop, &sset, fd);
+
+    // Get a service handle:
+    control_conn_t::handle_t h = find_service(fd, "test-service", service_state_t::STARTED,
+            service_state_t::STARTED);
+
+    // Issue a signal (SIGHUP for example):
+    std::vector<char> cmd = { DINIT_CP_SIGNAL, SIGHUP };
```
mobin-2008, line 1042: Hmm, I think I missing something but `SIGHUP` doesn't allocate 4 bytes which expected in `control.cc`. How it should be?

(Sorry I know I should fix anything related to this PR by myself but I think a simple question is reasonable)

davmac314: Since it's a vector of `char` the initialiser list is interpreted as a a list of `char` (1 byte each). The first `char` is `DINIT_CP_SIGNAL`, the next will be `SIGHUP`. To store `SIGHUP` as a sequence of four bytes you need to use the same technique as is used for a `handle_t` (which is also a four-byte integer) on the following lines. I.e. put it in a `int32_t` variable, copy each byte from the value into the vector.

davmac314: How is it that you're able to push commits to helican's repo btw?

mobin-2008: > How is it that you're able to push commits to helican's repo btw?

Because he enabled "Maintainers can edit this PR directly" which allow direct pushing for anyone have write access to main repo. (But only for PR's branch)

It's very useful thing for some situations like this.

src/tests/cptests/cptests.cc
```diff
@@ -951,6 +976,57 @@ void cptest_servicestatus()
...
+
+    bp_sys::supply_read_data(fd, std::move(cmd));
+    event_loop.regd_bidi_watchers[fd]->read_ready(event_loop, fd);
+
+    std::vector<char> wdata;
+    bp_sys::extract_written_data(fd, wdata);
+    assert(wdata.size() == 1);
+    assert(wdata[0] == DINIT_RP_ACK);
+
+    assert(bp_sys::last_sig_sent == SIGHUP);
```
davmac314, line 1023: This is basically a good idea for a test, but it is weak because it only checks one signal. The test could pass just because `bp_sys::last_sig_sent` happened to already be `SIGHUP` even if this wasn't changed at all by the control message. Ideally, it should now also send another DINIT_CP_SIGNAL now with a different signal number and check that `bp_sys::last_sig_sent` changes.

davmac314 (changes requested):
I've reviewed this whole thing again and there are a number of things that should be changed or fixed before I'm happy to include this:

- first, let's not make a habit of pushing changes to someone else's repo, even if we do have that ability. In cases like this I think the best thing is to pull the changes from their repo, push them to your own, and open a new PR. That's ok for this time though.
- it looks like "dinitctl signal (signal-name) --list" wouldn't error out, but it should. Correct syntax is either "dinitctl signal (signal-name) (service-name)" or "dinitctl signal --list". Rather than converting the signal name to a number where it currently happens (during arg parsing), it should be done in the argument post-processing (same place it does the special checks for ENABLE_SERVICE, DISABLE_SERVICE, SETENV, etc). Also the check for whether the signal is a number (using `std::stoi`) doesn't currently check if there is junk text after the number, it should (eg it currently would accept "99junk" and treat it as 99 instead of giving an error).
- let's rename `DINIT_RP_SIGNAL_BADPID`. It's being used when the service doesn't have a PID, but that's not the same as a "bad" PID. Maybe `DINIT_RP_SIGNAL_NO_PID`. And improve the error message in dinitctl by explaining that the service might not be a process service or it might be in the wrong state.
- `sig_num_t` type is defined as `int32_t` in two places. This is wrong, it should be `int`, and we don't need a typedef for this, just use the right type (int).
- signal number validation in control.cc is not necessary and should be removed. Instead the error value from `kill` (`bpsys::kill`) should be checked and if it is `EINVAL`, _then_ return the BADSIG error reply. The current code uses NSIG which isn't part of POSIX.
- can we prune the supported signal list (in `signal_to_int_map`) to a sensible set of signals that you might feasibly need to send to a process manually?
- there shouldn't be two versions of base_process_service_test that are doing the same thing, refactor it please.

Edit: some additional:
-  remove string_map_size template function, it's not needed - use sizeof instead (sizeof the whole array divided by size of one element gives you the number of elements). Also get rid of `string_to_int_map` struct and just use a `pair`. Initialisation of the array is also overly verbose, `string_to_int_map("none", 0)` could just be `{"none", 0}`.
- dinitctl man page: sentences should end at a newline

davmac314:
@mobin-2008  Some other small things:

- remove string_map_size template function, it's not needed - use `sizeof` instead (sizeof the whole array divided by size of one element gives you the number of elements). Also get rid of `string_to_int_map` struct and just use a `pair`. Initialisation of the array is also overly verbose, `string_to_int_map("none", 0)` could just be `{"none", 0}`.

src/includes/load-service.h
```diff
@@ -150,18 +150,100 @@ inline string_iterator skipws(string_iterator i, string_iterator end) noexcept
     return i;
 }
 
-// Convert a signal name to the corresponding signal number
-inline int signal_name_to_number(std::string &signame) noexcept
+// Utilities to map strings/signals to int
+// The signal list is created statically with only those signals,
+// which are defined for the current architecture/system
+std::pair<const char *, int> signal_to_int_map[] = {
```
mobin-2008, line 156: I wrote this draft new implementation and saw some weird stuff (mainly `ld` reported that `signal_to_int_map` duplicated multiple times). Seems like it's needed to be `inline`ed to work correctly but inline variables are C++17 specific thing. Needs more investigation.

@davmac314 Did you mean something like this?

davmac314: You've removed `constexpr`, that's why you're getting multiple definition errors.

davmac314: (other than that, yes, that's how the definition should look, but the signal list still needs to be trimmed down)

mobin-2008:
About signal list, I don't understand why we need to trim the list? For better memory use or because many of them have really rare use cases?

src/dinitctl.cc
```diff
@@ -324,8 +345,41 @@ int dinitctl_main(int argc, char **argv)
             cmdline_error = true;
         }
     }
+    else if (command == command_t::SIG_SEND && sig_num == -1) {
+        bool converted = false;
+        try
+        {
+            if (!std::all_of(sigstr.begin(), sigstr.end(), ::isdigit) &&
+                std::any_of(sigstr.begin(), sigstr.end(), ::isdigit)) {
```
mobin-2008, line 353: I'm not sure about this check, Are we have a signal which have a combine of numbers and characters?
and catching `std::bad_alloc` is enough?

mobin-2008:
> About signal list, I don't understand why we need to trim the list? For better memory use or because many of them have really rare use cases?

When I thought about it, it's reasonable to have a small list of predefined signals. My list include these:

- INT
- HUP
- USR1
- USR2
- KILL
- STOP
- CONT
- TERM

signals which I think reasonable. @davmac314 WDYT?

doc/manpages/dinitctl.8.m4
```diff
@@ -298,6 +301,17 @@ The value can be provided on the command line or retrieved from the environment
...
 Any subsequently started or restarted service will have these environment variables available.
 This is particularly useful for user services that need access to session information.
+.TP
+\fBsignal\fR
+Send a signal to a specified service.
+The \fIsignal-number\fR can either be specified as integer or signal name.
+Signal names are only defined for standard POSIX signals in abbrevated form
+(e.g. \fBHUP\fR, \fBKILL\fR, \fBUSR1\fR, \fBINT\fR, ...).
+Other system specific signals, like Linux real-time signals, have to be specified as integer.
+Dinit will accept singals betweeen \fB0\fR and \fBNSIG\fR.
```
davmac314, line 311: This text needs to be updated

src/dinitctl.cc
davmac314: Don't check it like this. What about USR1 and USR2?

`stoi` can tell you whether the input was completely converted. You'll need to pass the additional arguments to it.

I don't understand what would throw `bad_alloc` nor why you would catch it here.

src/dinitctl.cc
```diff
@@ -193,6 +200,10 @@ int dinitctl_main(int argc, char **argv)
                     break;
                 }
             }
+            else if (command == command_t::SIG_SEND && i == 2
```
davmac314, line 203: What is the `i == 2` check for?

src/includes/load-service.h
```diff
@@ -150,18 +150,100 @@ inline string_iterator skipws(string_iterator i, string_iterator end) noexcept
     return i;
 }
 
-// Convert a signal name to the corresponding signal number
-inline int signal_name_to_number(std::string &signame) noexcept
+// Utilities to map strings/signals to int
+// The signal list is created statically with only those signals,
+// which are defined for the current architecture/system
+const std::pair<const char *, int> signal_to_int_map[] = {
```
davmac314, line 156: Signal list has not been trimmed yet, why did you mark as ready for review?

davmac314 (changes requested):

src/includes/load-service.h
davmac314: Please don't request review until it is actually ready. If you have questions in the meantime you can ask them.

davmac314:
> When I thought about it, it's reasonable to have a small list of predefined signals. My list include these:

That looks pretty good to me. I would add `QUIT` and `INFO` (you need to check if SIGINFO is defined).

src/includes/load-service.h
mobin-2008: Sorry, I mean "Hey everything looks good and let's talk about signal list" but my opinion was changed; regardless than that, everything seems ok in my opinion.

src/dinitctl.cc
mobin-2008: For throw "unrecognised/invalid argument: --list" on `dinitctl signal <signal-name> --list` situation.

davmac314: Ok I guess you're trying to check that the command line is exactly "dinitctl signal --list". But:

- If I do "dinitctl signal KILL --list" -- It will say "--list" is an "unrecognized/invalid option" even though it is not. I just committed a fix for the same issue for other commands! I don't want to reintroduce the same bug now.
- what if I do: "dinitctl -u signal --list" for example? I don't think this should error (even though the `-u` doesn't really have any effect in this case).
- in general this is very brittle.

src/includes/load-service.h
davmac314: Until it's at the stage where you think that no more changes are needed, I'd prefer not to be reviewing it. Reviewing takes up a lot time (I know that's probably not obvious, but I have to be very careful and check a lot of things). So, please only request review when you think it's finished, and after you've carefully reviewed it yourself. If you're not totally sure that it's ok that's fine, but if there are definitely going to be changes needed then don't ask for review, instead ask questions to clarify those points.

src/dinitctl.cc
davmac314: > For throw "unrecognised/invalid argument: --list" on `dinitctl signal <signal-name> --list` situation.

Yeah but that's wrong. It's not an unrecognised/invalid argument. It's a recognized, valid argument that's given in the wrong context.

davmac314: I literally fixed this same issue just a few hours back:https://github.com/davmac314/dinit/commit/c3725a9e1dc8fe702efb737de6eeb0860516869f

I don't want to bring the same bug back now!

src/includes/load-service.h
davmac314: (So to be clear, if we need to talk about the signal list, that means it's not ready for review.)

mobin-2008:
Some cases I've been tested:
```bash
> src/dinitctl signal
dinitctl: signal number/name must be specified
!1> src/dinitctl signal --list
dinitctl: The following signal names are defined for this platform:
dinitctl: none   -> 0   NONE   -> 0   HUP    -> 1   INT    -> 2   
dinitctl: QUIT   -> 3   KILL   -> 9   USR1   -> 10  USR2   -> 12  
dinitctl: TERM   -> 15  CONT   -> 18  STOP   -> 19  
> src/dinitctl signal HUP
dinitctl: service name must be specified
!1> src/dinitctl -p ./sock signal HUP boot
dinit: Service boot terminated due to signal 1
[STOPPD] boot
[  OK  ] boot
> src/dinitctl -p ./sock signal 36 boot # SIGRTMIN+1 on Linux-x86_64
dinit: Service boot terminated due to signal 36
[STOPPD] boot
[  OK  ] boot
> src/dinitctl -p ./sock signal 332109 boot
dinit: Requested signal not in valid signal range.
dinitctl: provided signal was invalid.
!1> src/dinitctl -p ./sock signal junk99 boot
dinitctl: 'junk99' is not a valid signal name/number
!1> src/dinitctl -p ./sock signal 99junk boot
dinitctl: '99junk' is not a valid signal name/number
!1> src/dinitctl -p ./sock signal HUP --list
dinitctl: Invalid command line.
Try 'dinitctl --help' for more information.
!1> src/dinitctl -p ./sock signal --list HUP
dinitctl: Invalid command line.
Try 'dinitctl --help' for more information.
```

davmac314:
> Some cases I tested it:

Thanks. I think we need to remove "none" and "NONE" from the list of signals displayed.

Have you tried:
```
dinitctl signal INT boot other
```
?

mobin-2008:
> 

Seems like it tries to find "other". Are we need to fix it?

src/dinitctl.cc
```diff
@@ -1813,6 +1886,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
...
+    }
+    rbuffer.consume(1);
+    return 0;
+}
+
+static int signal_list()
+{
+    using namespace std;
+    using namespace dinit_load;
+    cerr << "dinitctl: The following signal names are defined for this platform:";
```
davmac314, line 1936: Should be `cout` not `cerr`. It's not an error.

src/dinitctl.cc
```diff
@@ -1813,6 +1886,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
...
+    int i = 0;
+    for (const auto &signal: signal_to_int_map) {
+        if (i%4 == 0) {
```
davmac314, line 1947: Please use style consistent with the rest of the code base, including spaces around operators, indentation.

src/dinitctl.cc
```diff
@@ -1813,6 +1886,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
...
+        cerr << endl << "dinitctl: ";
+        }
+        string sigpad = signal.first;
+        string numpad = to_string(signal.second);
+        sigpad.resize(7,' ');
+        numpad.resize(4,' ');
+        cerr << sigpad << "-> " << numpad;
```
davmac314, line 1954: should be `cout`

src/dinitctl.cc
```diff
@@ -1813,6 +1886,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
...
+        i++;
+    }
+    cerr << endl;
```
davmac314, line 1957: `cout`

src/includes/load-service.h
```diff
@@ -150,18 +150,36 @@ inline string_iterator skipws(string_iterator i, string_iterator end) noexcept
     return i;
 }
 
-// Convert a signal name to the corresponding signal number
-inline int signal_name_to_number(std::string &signame) noexcept
+// Utilities to map strings/signals to int
+// The signal list is created statically with only those signals,
```
davmac314, line 154: Comment is outdated

doc/manpages/dinitctl.8.m4
```diff
@@ -298,6 +301,18 @@ The value can be provided on the command line or retrieved from the environment
 called in.
 Any subsequently started or restarted service will have these environment variables available.
 This is particularly useful for user services that need access to session information.
+.TP
+\fBsignal\fR
+Send a signal to a specified service.
+The \fIsignal-number\fR can either be specified as integer or signal name.
```
davmac314, line 307: "specifed as _an_ integer or signal name". Don't worry about this though, I will fix the grammar once it's merged (and probably change the wording a little).

doc/manpages/dinitctl.8.m4
```diff
@@ -298,6 +301,18 @@ The value can be provided on the command line or retrieved from the environment
...
+Signal names are only defined for these signals:
+\fBHUP\fR, \fBKILL\fR, \fBUSR1\fR, \fBUSR2\fR, \fBSTOP\fR, \fBINT\fR, \fBTERM\fR,
+\fBCONT\fR, \fBQUIT\fR and if system supports it; \fBINFO\fR.
+Other system specific signals, like Linux real-time signals, have to be specified as integer.
+Dinit will accept any signal which supported by \fBkill(2)\fR syscall.
+.sp
+The \fB--list/\-l\fR option can be used (instead of \fIsignal-number\fR and \fIservice-name\fR)
```
davmac314, line 314: "--list/-l" will all be in bold, it's probably better to say "--list or -l" or at least make the slash regular font instead of bold. I will fix it later.

src/dinitctl.cc
```diff
@@ -304,6 +323,14 @@ int dinitctl_main(int argc, char **argv)
                 }
                 to_service_name = argv[i];
             }
+            else if (command == command_t::SIG_SEND) {
+                if (!show_siglist) {
+                    if (sigstr.empty()) sigstr = argv[i]; else service_name = argv[i];
```
davmac314: Please follow style.

src/dinitctl.cc
```diff
@@ -324,8 +351,47 @@ int dinitctl_main(int argc, char **argv)
             cmdline_error = true;
         }
     }
+    else if (command == command_t::SIG_SEND && sig_num == -1) {
```
davmac314, line 362: I don't think `sig_num == -1` could be false? Isn't it redundant?

src/dinitctl.cc
```diff
@@ -324,8 +351,47 @@ int dinitctl_main(int argc, char **argv)
...
+            if (sig_num == -1) {
+                try
+                {
+                    size_t pos;
+                    sig_num = std::stoi(sigstr, &pos);
+                    if (sigstr.size() != pos) {
+                        throw std::invalid_argument("");
+                    }
+                }
+                catch (std::invalid_argument)
```
davmac314, line 390: Is this the only exception that needs to be caught?

src/dinitctl.cc
```diff
@@ -324,8 +351,47 @@ int dinitctl_main(int argc, char **argv)
...
+        else {
+            if (sigstr.empty()) {
+                cerr << "dinitctl: signal number/name must be specified" << std::endl;
+                return 1;
+            }
+            if (service_name == nullptr) {
+                cerr << "dinitctl: service name must be specified" << std::endl;
+                return 1;
+            }
+            sig_num = dinit_load::signal_name_to_number(sigstr);
```
davmac314, line 380: If user specifies signal "none"/"NONE" there should be an error.

davmac314 (changes requested):
It's getting there, but still some issues. Also need to fix the error cases noted in the PR discussion.

src/includes/load-service.h
mobin-2008: What do you mean? I changed it a bit.

src/dinitctl.cc
```diff
@@ -324,8 +359,52 @@ int dinitctl_main(int argc, char **argv)
...
+            else {
+                cmdline_error = true;
+            }
+        }
+        else {
+            if (sigstr.empty()) {
+                cerr << "dinitctl: signal number/name must be specified" << std::endl;
+                return 1;
+            }
+            else if (sigstr == "NONE" | sigstr == "none") {
```
davmac314, line 376: It's not a good idea to use constants (like `"NONE"` and `"none"`) in multiple places without creating a named constant (variable). Because, if it gets changed in one place, there is no indication that it needs to be changed in the other place too.

There's no need to refer to "none" and "NONE" here though, we can check if `signal_name_to_number` returns 0 instead. That solves the problem much more cleanly.

src/dinitctl.cc
```diff
@@ -324,8 +359,52 @@ int dinitctl_main(int argc, char **argv)
...
+            if (sig_num == -1) {
+                try
+                {
+                    size_t pos;
+                    sig_num = std::stoi(sigstr, &pos);
+                    if (sigstr.size() != pos) {
+                        throw std::invalid_argument("");
+                    }
+                }
+                catch (std::exception)
```
davmac314, line 395: Ok, but at least add a comment saying what exception types this is expected to handle.

Someone reading this code should be able to look at it and think "Ok, this is designed to handle both invalid_argument and out_of_range".

src/dinitctl.cc
```diff
@@ -484,6 +564,12 @@ int dinitctl_main(int argc, char **argv)
             }
             return cat_service_log(socknum, rbuffer, service_name, catlog_clear);
         }
+        else if (command == command_t::SIG_SEND) {
+            return signal_send(socknum, rbuffer, service_name, sig_num);
+        }
+        else if (command == command_t::SIG_LIST) {
+            return signal_list();
```
davmac314, line 571: This doesn't need a control socket connection, but one has been opened anyway. That should really be fixed. I.e. don't open a control socket for a command that doesn't need it.

davmac314 (changes requested):

src/dinitctl.cc
```diff
@@ -324,8 +359,52 @@ int dinitctl_main(int argc, char **argv)
...
+            }
+            sig_num = dinit_load::signal_name_to_number(sigstr);
+            if (sig_num == 0) {
+                cerr << "dinitctl: '" << sigstr
+                    << "' is not a valid signal name/number" << std::endl;
+                return 1;
+            }
+            else if (sig_num == -1) {
+                try
+                {
```
davmac314, line 388: Please match style used elsewhere.

src/dinitctl.cc
```diff
@@ -401,6 +481,11 @@ int dinitctl_main(int argc, char **argv)
         return 1;
     }
     
+    // SIG_LIST doesn't need a control socket connection so Handle it specially.
```
davmac314, line 484: Good comment, but "Handle" doesn't need a capital H

src/dinitctl.cc
```diff
@@ -1813,6 +1901,70 @@ static int trigger_service(int socknum, cpbuffer_t &rbuffer, const char *service
...
+    }
+    rbuffer.consume(1);
+    return 0;
+}
+
+static int signal_list()
+{
+    using namespace std;
+    using namespace dinit_load;
+    cout << "dinitctl: The following signal names are defined for this platform:";
```
davmac314, line 1951: Change this to "... are supported" instead of "... are defined for this platform"

src/control.cc
```diff
@@ -987,6 +990,53 @@ bool control_conn_t::process_catlog()
...
+    // Reply:
+    // 1 byte packet type = DINIT_RP_*
+
+    pid_t spid = service->get_pid();
+    // we probably don't want to kill/signal every process (in the current group),
+    // but get_pid() sometimes returns -1 if e.g. service is not 'started'
+    if (spid == -1 || spid == 0) {
+        char nak_rep[] = { DINIT_RP_SIGNAL_NOPID };
+        return queue_packet(nak_rep, 1);
+    } else {
```
davmac314, line 1024: style.

src/dinitctl.cc
```diff
@@ -324,8 +359,52 @@ int dinitctl_main(int argc, char **argv)
...
+                    size_t pos;
+                    sig_num = std::stoi(sigstr, &pos);
+                    if (sigstr.size() != pos) {
+                        throw std::invalid_argument("");
+                    }
+                }
+                catch (std::exception) // catch both invalid_argument and out_of_range
+                {
+                    cerr << "dinitctl: '" << sigstr
+                        << "' is not a valid signal name/number" << std::endl;
```
davmac314, line 398: Style. "When wrapping a line, the indent should be increased by at least two levels on the subsequent lines"

davmac314: (i.e. same as is done elsewhere).

davmac314 (approved):
This is great, thanks!

mobin-2008:
Closes #134

james-knippes:
> Hi @james-knippes - just wondering if you were still planning on finishing off this PR? I think that tests are the only thing pending

Hey, sorry for taking so long. Got distracted by life stuff. Should have responded, that I did not have the time.
Thank you @mobin-2008 for taking over. Nice work in general and also with the test (:

davmac314:
@james-knippes no problem, thanks for the work that you did on this.

PR #182, opened 2023-05-06
make, meson: Some configurable options in compile time

### LINKED ISSUE #180: make, meson: Some configurable options in compile time
Hi. #172 changed default behavior of `restart` option. I thought if someone doesn't want something like that, What he must do? Probably he needs to change this line in `include/load-service.h`:
https://github.com/davmac314/dinit/blob/0d6e509a1d14191ecb8cd4374e01c16f82c2a942/src/includes/load-service.h#L817

but I think it's must have a configurable option in `mconfig`(s), `configure` and meson builds.

So I think these options are useful:
```sh
DEFAULT_AUTO_START=true
DEFAULT_START_TIMEOUT=60 # For some faster/slower systems
DEFAULT_STOP_TIMEOUT=10 # For some faster/slower systems
```
I do these changes, This issue has created for tracking its process.

Regards

mobin-2008:
Hi.
DEFAULT_AUTO_RESTART, DEFAULT_START_TIMEOUT and DEFAULT_STOP_TIMEOUT are new options available in compile time for configuring some default options. Default values are in brackets:

DEFAULT_AUTO_RESTART: Enable/disable auto-restart for services by default [true]
DEFAULT_START_TIMEOUT: Default start-timeout for services [60]
DEFAULT_STOP_TIMEOUT: Default stop-timeout for services [10]

## ToDo
- [x] Add those options to mconfigs and configure
- [ ] Add those options to meson
- [ ] support nano-seconds in timeouts options
- [ ] Apply options values to man-pages

Regards

Closes #180


mobin-2008:
I working on those options and I think make based build system mostly ready for those but meson needs more work. Am I need to add some tests about those to Unit tests? Uhhh, I think no.

mobin-2008:
It's ready to be reviewed. I gonna mark some lines for more discussions.

build/tools/mconfig-gen.cc
```diff
@@ -52,6 +53,47 @@ void parse_arg(std::string arg)
...
+    auto len = paramval.length();
+    decltype(len) i;
+    for (i = 0; i < len; i++) {
+        char ch = paramval[i];
+        if (ch == '.' || ch == ',') {
+            i++;
+            break;
+        }
+        if (ch < '0' || ch > '9') {
+            std::cerr << "mconfig-gen: " << std::string("bad value for ") + paramname << "\n";
```
mobin-2008, line 72: C++ exceptions can be used?

doc/manpages/dinit-service.5.m4
```diff
@@ -168,7 +168,7 @@ That means this path is static and cannot itself have variable substitutions (se
 Indicates whether the service should automatically restart if it stops, including due to
 unexpected process termination or a dependency stopping.
 Note that if a service stops due to user request, automatic restart is inhibited.
-The default is to automatically restart.
+The default is to restart=$$$DEFAULT_AUTO_RESTART@@@.
```
mobin-2008, line 171: I changed its words but it was better before. I didn't find better words to integrate with that variable.

meson.build
```diff
@@ -6,7 +6,7 @@ project(
         ['sh', 'build/general_variables_for_meson.sh', 'version'],
         check : true).stdout(),
     license : 'Apache-2.0',
-    meson_version : '>= 0.47.0',
+    meson_version : '>= 1.0.0',
```
mobin-2008, line 9: Meson version required version bumped due to `'in' operator for string variables`. Is it worth its backward compatibility breaking? I think yes. except if a better solution found.

Those are all of my points about this PR.

meson.build
```diff
@@ -55,6 +61,21 @@ mconfig_data.set('DINIT_VERSION', '"' + meson.project_version() + '"')
 mconfig_data.set('SYSCONTROLSOCKET', '"' + dinit_control_socket_path + '"')
 mconfig_data.set('SBINDIR', '"' + sbindir + '"')
 mconfig_data.set('SHUTDOWN_PREFIX', '"' + shutdown_prefix + '"')
+mconfig_data.set('DEFAULT_AUTO_RESTART', default_auto_restart)
+if '.' in default_start_timeout
```
davmac314, line 65: This is clunky (checking for '.' and handling it differently in either case, plus repeating that for several different things). Can this be simplified (even by dropping support for sub-second times?)

Also, I just realised meson builds aren't useing mconfig-gen to generate mconfig.h. I think that it would be better to use mconfig-gen; it can do a lot more sanity checking a lot easier, and it generates constants where appropriate instead of always `#define`'s.

davmac314 (changes requested):

meson.build
davmac314: But you should update documentation accordingly

meson.build
mobin-2008: Hmm, I think nobody use sub-second times in default times, so it's reasonable to remove it.

About meson, I like to talk about it later (after resolving some other things).

doc/manpages/dinit-service.5.m4
mobin-2008: Uhh, seems like `ifelse()` isn't processed in our .m4 files. man pages just print `ifelse(DEFAULT_AUTO_RESTART, true, The default is to automatically restart, The default is to not automatically restart)` but it works in my m4 interactive shell.

davmac314: If it works in the interactive shell, that should mean that it works generally...

Perhaps you're not (un)quoting it right? It needs `$$$` before the entire construct and `@@@` afterwards, something like

```
$$$ifelse(DEFAULT_AUTO_RESTART, true, @@@The default is to automatically restart$$$, @@@The default is to not automatically restart$$$)@@@
```

mobin-2008: I thought `$$$` is for resolving variables but it doesn't, interesting. I changed it to:
```m4
$$$ifelse(DEFAULT_AUTO_RESTART, true, The default is to automatically restart.,
    The default is to not automatically restart.)
@@@
```
Seems OK.

BUILD
```diff
@@ -122,6 +122,19 @@ USE_UTMPX=1|0
     like "who" to work correctly (the service configuration items "inittab-id" and "inittab-line"
     have no effect if this is disabled). If not set to any value, support is enabled for certain
     systems automatically and disabled for all others.
+USE_INITGROUPS=1|0
```
davmac314, line 125: This change shouldn't be part of this merge req...

davmac314: Seems like you squashed other commit(s) into this one, you'll need to undo that

meson_options.txt
```diff
@@ -52,6 +58,24 @@ option(
...
+    'default-auto-restart',
+    type : 'boolean',
+    value : true,
+    description : 'Enable/disable auto-restart for services by default.'
+)
+option(
+    'default-start-timeout',
+    type : 'integer',
+    value : 60,
+    description : 'Default start-timeout for services'
```
davmac314, line 71: please specify unit (seconds)

BUILD
davmac314: The commit still contains the change.

davmac314: Also, don't merge master into your PR branches. Rebase your PR branch onto master instead. That will probably fix this anyway.

configure
```diff
@@ -139,8 +139,15 @@ Optional options:
   --disable-cgroups             Disable Cgroups support
   --enable-utmpx                Enable manipulating the utmp/utmpx database via the related POSIX functions [auto]
   --disable-utmpx               Disable manipulating the utmp/utmpx database via the related POSIX functions
+<<<<<<< HEAD
```
davmac314, line 142: I think you forgot to resolve this conflict

meson.build
```diff
@@ -33,7 +33,13 @@ fuzzer = get_option('fuzzer')
 man_pages = get_option('man-pages')
 support_cgroups = get_option('support-cgroups')
 use_utmpx = get_option('use-utmpx')
+<<<<<<< HEAD
```
davmac314, line 36: Another conflict that needs to be resolved

meson.build
```diff
@@ -33,7 +33,9 @@ fuzzer = get_option('fuzzer')
 man_pages = get_option('man-pages')
 support_cgroups = get_option('support-cgroups')
 use_utmpx = get_option('use-utmpx')
-use_initgroups = get_option('use-initgroups')
```
mobin-2008, line 36: I should add these lines marked as "removed"? I think should.

davmac314 (approved):

PR #184, opened 2023-05-15
meson: Remove workaround for `version.conf`

mobin-2008:
And replace it with grep + .substring() based solution.


meson.build
```diff
@@ -2,9 +2,8 @@
 project(
     'dinit', 
     'cpp', 
-    version : run_command(
-        ['sh', 'build/general_variables_for_meson.sh', 'version'],
-        check : true).stdout(),
+    version : run_command('grep', 'VERSION', 'build/version.conf',
```
davmac314, line 5: This technique is fragile. I think it should at least:

-  look for `VERSION=` instead of just `VERSION` (and similar for other variables). And require that the string is found at the beginning of the line.  This is more robust in case new variables are added later that begin with or contain the same letters.
- use the calculated length of `VERSION=` instead of hardcoding the value (8). Though, meson doesn't seem to have a documented way to determine the length of a string. Perhaps `len('VERSION=')` will work. Reason: if it is hardcoded it is easy to miss updating it in case the name changes.

davmac314 (changes requested):

meson.build
eli-schwartz: What about `.stdout().strip().split('=')[1]`? Split it into an array on each equals sign, and take the second component. You know there will only ever be two components so you don't need to assert that.

mobin-2008: I replaced `.substring()` with `.split('=')` for more robust variable definition. it's fix hardcoded `.substring()` for finding variable values.
But
> And require that the string is found at the beginning of the line.

Hmm, About this I didn't found a robust solution. but I found some ideas:
1. `grep` support ending up finding paterns via `-m` flag after found n matching numbers of line. for example if `VERSION=` before `TEST_VERSION=`, we can use `grep -m1 VERSION= build/version.conf`.

2. meson have `keyval` module designed for reading these files. it's robustest way to do that but modules in meson can't be called before `project()` function and I think it means we can't use it for version. for example
    ```meson
    ...
    meson_version : '>= 0.56.0'
    ...

    ## Import variables from build/version.conf
    kv = import('keyval')
    version_conf = kv.import('build/version.conf`)

   ## General defines
   ...
   month = version_conf.get('MONTH')
   year = version_conf.get('YEAR')
   ...
   ```
Regards

davmac314: >  >   And require that the string is found at the beginning of the line.

> Hmm, About this I didn't found a robust solution

How about "grep ^VERSION= build/version.conf"?

BUILD_MESON
```diff
@@ -57,8 +57,7 @@ Everything is ready! you can compile Dinit with
      meson compile -C builddir
 
 This command builds the dinit, dinitctl and optionaly the shutdown/reboot/halt
-utilitys by default. Note: on Mesons older than 0.54, you need to use
-"ninja -C builddir" instead of "meson compile -C builddir".
+utilitys by default.
```
davmac314, line 60: "utilitys" -> "utilities"

BUILD_MESON
```diff
@@ -209,7 +208,7 @@ Why we use another option for sbindir?
 
 For some reasons, Old Mesons force sbindir will prefixed by Meson's prefix
 (eg /usr + sbin) but Dinit project use /sbin by default as sbindir.
-in Newer Mesons that fixes this thing but we target Mesons's 0.47.0 as
+in Newer Mesons that fixes this thing but we target Mesons's 0.56.0 as
```
davmac314, line 211: What version of Meson is this actually fixed in?

eli-schwartz: https://mesonbuild.com/Release-notes-for-0-62-0.html#all-directory-options-now-support-paths-outside-of-prefix

mobin-2008: Hmm, it's a bit werid but no one talks about that issue and its fix in release notes (or just I didn't see something related to that)! related fix merged in 9 Feb 2022 so it's fixed somewhere at release 0.6x.x.

davmac314: Can we change the note here to explicitly mention the version (0.62.0 according to Eli's post)?

mobin-2008: Looks ok, I will do that.

mobin-2008:
Everything looks good. I Just wait for Meson CI:
- [ ] Debian ()
- [ ] FreeBSD ()
- [ ] OpenBSD ()
- [ ] Alpine ()

davmac314 (approved):

PR #185, opened 2023-05-16
baseproc-service: fix default log target for none-logged services

q66:
We need to treat NONE (which is the default) the same as LOGFILE, so that we get a default /dev/null target. Otherwise, it is treated as BUFFER by default, which is undesirable, as processes will suddenly start hanging as soon as they run out of buffer space.

### DIFF
--- a/src/baseproc-service.cc
+++ b/src/baseproc-service.cc
@@ -106,7 +106,7 @@ bool base_process_service::start_ps_process(const std::vector<const char *> &cmd
     }
 
     const char * logfile = this->logfile.c_str();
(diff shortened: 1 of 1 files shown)


davmac314 (approved):

q66:
btw, have you considered having a "syslog" target to automatically redirect service stdout/err to syslog?

davmac314:
Yes - thinking about something like that by pushing messages through dinit's own log output buffer. Also, I'd like to be able to log via another process (managed as a separate service).

Issue #151, opened 2023-01-28
run-as user gropups


kopavel:
If service started with run-as option process don't get all user groups - only basic one.
As alternative i pushed not to use "run-as" option but instead:
   command = su - <username> -c "service command" 
but it's no so clear.

davmac314:
This is the documented behaviour (from `dinit-service(5)`):

> If specified by name, the
              group for the process will also be set to the primary group of
              the specified user

Implementing the addition of supplementary groups requires iterating through all the groups to see which one claims the user as a member. This is probably not difficult, but it's also an unnecessary overhead in most cases.

What is your use case such that this is needed?

kopavel:
I make minimalistic kiosk-mode system for not so new hardware. When OpenBox starts already with system without any DesktopManager like LigthDM or any other. Start it like Dinit  service as unprivileged user with dependency from Xorg.  And that user doesn't have any additional group, such a AUDIO for example. Yes - not really correct solution. But really minimalistic - only 100Mb ram usage with GUI and about 10 sec start time. For now i start like 
          "waitXorgSocket.sh;su -u username -c ";export DISPLAY=:0.0;openbox-session"  
and all work. Only have shell process (su) remaining  between INIT and Openbox. 
 waitXorgSocket.sh  - shell in witch I wait X to provide socket when it's really start. In other way OpenBox can't start and fall several times. (Xorg process fork to background doesn't mean Xorg are ready to work - need to wait till its open DISPLAY for connection.)

q66:
this should really be as simple as this:

```
--- a/src/run-child-proc.cc
+++ b/src/run-child-proc.cc
@@ -332,6 +332,12 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
         err.stage = exec_stage::SET_UIDGID;
         // We must set group first (i.e. before we drop privileges)
         if (setregid(gid, gid) != 0) goto failure_out;
+        errno = 0;
+        auto *pw = getpwuid(uid);
+        if (pw) {
+            if (initgroups(pw->pw_name, gid) != 0) goto failure_out;
+        }
+        else if (errno) goto failure_out;
         if (setreuid(uid, uid) != 0) goto failure_out;
     }
 
```

The interface is not POSIX, but as far as I can tell it's present on every system worth supporting (it works on Linux, every BSD, Solaris, macOS, and possibly others)

PR #187, opened 2023-05-17
run-child-proc: initialize supplementary groups for named run-as

q66:
This is practically portable (exists on Linux, BSDs, Solaris, etc) despite not being POSIX, so enable it unconditionally without extra checks. It can still be disabled explicitly if needed.

Fixes https://github.com/davmac314/dinit/issues/151

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -63,6 +63,11 @@ if use_utmpx.enabled()
 elif use_utmpx.disabled()
     mconfig_data.set('USE_UTMPX', '0')
 endif
+if use_initgroups.enabled()
+    mconfig_data.set('USE_INITGROUPS', '1')
+elif use_initgroups.disabled()
+    mconfig_data.set('USE_INITGROUPS', '0')
+endif
 
 ## Outputs
 subdir('src')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -41,6 +41,12 @@ option(
     description : 'Whether to build support for manipulating the utmp/utmpx database via the related POSIX functions.'
 )
 option(
+    'use-initgroups',
+    type : 'feature',
+    value : 'auto',
+    description : 'Whether to use initgroups to initialize supplementary groups for run-as services.'
+)
+option(
     'dinit-sbindir',
     type : 'string',
     value : '/sbin',
--- a/BUILD
+++ b/BUILD
@@ -122,6 +122,10 @@ USE_UTMPX=1|0
     like "who" to work correctly (the service configuration items "inittab-id" and "inittab-line"
     have no effect if this is disabled). If not set to any value, support is enabled for certain
     systems automatically and disabled for all others.
+USE_INITGROUPS=1|0
+    Whether to initialize supplementary groups for run-as services. The C API for this is not
+    in POSIX, but is in practice supported on just about every relevant system, so it is enabled
+    by default. If it is not supported on yours, you can explicitly disable it.
 
 
 Running test suite
--- a/BUILD_MESON
+++ b/BUILD_MESON
@@ -118,6 +118,16 @@ Custom options:
                              Available values : enabled, disabled, auto
                              Default value : auto
 
+ use-initgroups            : Whether to use initgroups to initialize supplementary
+                             groups in services that use run-as with a name. Without
+                             this the services may potentially misbehave as process
+                             and its child processes won't recognize being in the
+                             supplementary groups, only their primary group. However,
+                             some systems may not implement the initgroups API, so
+                             it may be necessary to disable it on those.
+                             Available vlaues : enabled, disabled, auto
+                             Default value : auto
+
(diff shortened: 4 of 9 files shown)


meson.build
```diff
@@ -63,6 +63,11 @@ if use_utmpx.enabled()
 elif use_utmpx.disabled()
     mconfig_data.set('USE_UTMPX', '0')
 endif
+if use_initgroups.enabled()
+    mconfig_data.set('USE_INITGROUPS', '1')
+elif use_initgroups.disabled()
+    mconfig_data.set('USE_INITGROUPS', '0')
+endif
```
mobin-2008, line 70: I know `grp.h` has been defined by UNIX single spec and also it's available on most systems but looks good to check that `grp.h` is available or not with these simple command:
```meson
elif use_initgroups.auto() and compiler.check_header('grp.h')
    mconfig_data.set('USE_INITGROUPS', '1')
endif
```

meson_options.txt
```diff
@@ -40,6 +40,12 @@ option(
     value : 'auto',
     description : 'Whether to build support for manipulating the utmp/utmpx database via the related POSIX functions.'
 )
+option(
```
mobin-2008, line 43: Thanks for adding new option to meson. Please also add it to `MESON_BUILD` doc.

meson.build
davmac314: @mobin-2008 it is guaranteed by POSIX that grp.h exists. Since we target POSIX systems explicitly, and since we already require this header to build anyway, we don't need to check for the header.

But more importantly, even if the header is there, that doesn't mean that `initgroups` will be defined. Setting `USE_INITGROUPS` based on whether the header exists will just cause it to be set on all systems that Dinit can be built on.

@q66:
Assuming the function will be available makes me uneasy - I've previously built Dinit successfully on systems like Sortix, which doesn't have initgroups - but I agree it's almost universal, so I'm ok with expecting it to be available at least for now.

However, there is still a good reason for not initialising the groups especially on system-level services. Specifically, intiialising group list based on a particular user id involves iterating the entire set of groups to find each group which lists the user as a member. While this isn't so bad for a single-user system where the groups list is entirely local, it is entirely possible that the groups database is read over the network (via NIS). (With Glibc this is configured via /etc/nsswitch.conf). We really don't want this happening during startup for essential services before the network is properly set up.

So, I think initialising the supplementary groups should be a service option, not a default.

davmac314 (changes requested):

meson.build
q66: IMO it should be default, considering 1) run-as services aren't that common in the first place 2) the case you are talking about is rare, usage of dinit with it is going to be even rarer (i'd be surprised if there was even one user like that), and even then I doubt that the overhead is going to be meaningful, and if visible at all, detrimental to service startup time 3) an environment without supplementary groups initialized will run into all sorts of potential subtle misbehaviors

davmac314: It's a common setup in institutional/corporate environments (and it can definitely cause noticeable slowdowns) but I accept that in current typical setups where Dinit is actually used it won't be an issue, and ignoring supplementary groups is not great. So, ok, you've convinced me. I still think there should be an option not to set the supplementary groups, but I can add that at a later point.

q66: well, one thing is the slowdown being noticeable, the other is whether it actually matters - most system services that run as another user are capable of dropping privileges by themselves (and they choose themselves if auxiliary groups are needed or not), and those that do not - to cause a meaningful slowdown, you would need a 1) run-as service that is necessary to start before login 2) ideally have multiple of them 3) the group initialization would have to take at least a few seconds

unless those conditions are met, the worst case is that your boot may be slowed down by up to a couple seconds in worst case (and typically even that won't happen) which is IMO a reasonable tradeoff - I've always been of the opinion that when given the choice between maximum performance and robustness, robustness should win unless it is proven that the performance is actually necessary

if the performance becomes an actual problem in a real scenario, we could always add a service option to prevent the group initialization, which would be fine because then the service would have to acknowledge "i know what i am doing"

davmac314: >  you would need a 1) run-as service that is necessary to start before login 2) ideally have multiple of them 3) the group initialization would have to take at least a few seconds

True, though (3) is definitely plausible and (2) isn't really necessary when the network is crawling. It would be a fair point that in that situation nobody's going to be able to log in anyway, though.

>  I've always been of the opinion that when given the choice between maximum performance and robustness, robustness should win unless it is proven that the performance is actually necessary

This is in essence why I agreed:

> So, ok, you've convinced me

meson_options.txt
davmac314: This is the only thing outstanding before merging I think.

q66: alright, min

q66: done

davmac314: Merged! Thanks

PR #190, opened 2023-05-24
configure, meson: Always define `USE_INITGROUPS`

mobin-2008:
I think it is better now.


BUILD_MESON
```diff
@@ -124,8 +124,8 @@ Custom options:
                              supplementary groups, only their primary group. However,
                              some systems may not implement the initgroups API, so
                              it may be necessary to disable it on those.
-                             Available vlaues : enabled, disabled, auto
-                             Default value : auto
+                             Available vlaues : true, false
```
davmac314, line 127: There's a (pre-existing) typo: "Available vlaues" -> "Available values"

davmac314 (changes requested):
Looks good if you can fix the typo, thanks

davmac314 (approved):

PR #192, opened 2023-05-27
CI: Meson CI(s) on GitHub.

mobin-2008:
Previously Meson based CI hosted on SourceHut. That was great but sr.ht doesn't have good support for integrating with GitHub due to sunset of 'dispatch.sr.ht`. So I moving my workflows from there to GitHub for better Integration.

These are designed to be in main repo (davmac314/dinit) but these aren't running by default. These will be ran when a PR in main repo has "meson" label.

Regards

### DIFF
--- a/.github/workflows/meson_ci.yml
+++ b/.github/workflows/meson_ci.yml
@@ -29,6 +29,9 @@ on:
   pull_request:
     types:
       - labeled
+      - opened
+      - reopened
+      - synchronize
     branches:
       - master
     paths:
--- a/.github/workflows/meson_ci.yml
+++ b/.github/workflows/meson_ci.yml
@@ -1,0 +1,149 @@
+## This Workflow will run when a PR has "meson" label or when something pushed to master.
+
+name: Dinit Meson based CI
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '**'
+      - 'build/version.conf'
+      - '!.*'
+      - '!README.md'
+      - '!NEWS'
+      - '!BUILD*'
+      - '!CONTRIBUTORS'
+      - '!TODO'
+      - '!LICENSE'
+      - '!**/Makefile'
+      - '!build/**'
+      - '!doc/CODE-STYLE'
+      - '!doc/COMPARISON'
+      - '!doc/CONTRIBUTING'
+      - '!doc/DESIGN'
+      - '!doc/getting_started.md'
+      - '!doc/manpages/README'
+      - '!doc/manpages/generate-html.sh'
+      - '!doc/linux/**'
+  pull_request:
+    types:
+      - labeled
+    branches:
+      - master
+    paths:
+      - '**'
+      - 'build/version.conf'
+      - '!.*'
+      - '!README.md'
+      - '!NEWS'
+      - '!BUILD*'
+      - '!CONTRIBUTORS'
+      - '!TODO'
+      - '!LICENSE'
+      - '!**/Makefile'
+      - '!build/**'
+      - '!doc/CODE-STYLE'
+      - '!doc/COMPARISON'
+      - '!doc/CONTRIBUTING'
+      - '!doc/DESIGN'
+      - '!doc/getting_started.md'
+      - '!doc/manpages/README'
+      - '!doc/manpages/generate-html.sh'
+      - '!doc/linux/**'
+  workflow_dispatch:
+
+jobs:
+
+  Debian-bullseye_build:
+
+    if: ${{ github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'meson') }}
+    runs-on: ubuntu-latest
+    container:
+      image: debian:bullseye
+    strategy:
+      matrix:
+        include:
+          - arch: 'amd64'
+          - arch: 'i386'
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Add i386 repos
+      if: ${{ matrix.arch == 'i386' }}
+      run: dpkg --add-architecture i386
+    - name: Getting depends (amd64)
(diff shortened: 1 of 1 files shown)


mobin-2008:
As see, Those just skipped.

mobin-2008:
Hmm, seems like It's need a another commit after assign "meson" label :/

davmac314:
Since `workflow-dispatch` is a trigger, it can be run manually if needed (i.e. for new pull requests where the label isn't set yet).

mobin-2008:
> Since `workflow-dispatch` is a trigger, it can be run manually if needed (i.e. for new pull requests where the label isn't set yet).

`workflow-dispatch` is for local branches, It doesn't work for PR(s).

davmac314:
Ah, you're right. That's an annoying limitation.

mobin-2008:
I changed code to run meson workflows when a "rerun" requested. I think it's ready now.

.github/workflows/meson_ci.yml
```diff
@@ -0,0 +1,124 @@
...
+      - '!doc/manpages/README'
+      - '!doc/manpages/generate-html.sh'
+      - '!doc/linux/**'
+  workflow_dispatch:
+
+jobs:
+
+  Debian-bullseye_build:
+
+    if: ${{ github.run_attempt >= 2 || contains(github.event.pull_request.labels.*.name, 'meson') }}
```
davmac314, line 35: So *any* rerun of CI will always run the meson CI builds? Is that what you intended?

davmac314: Could we just leave the condition as it was ("pull request label contains 'meson'") and then if we want to run meson CI for a given pull request just add the label and then re-run CI?

mobin-2008: > So _any_ rerun of CI will always run the meson CI builds? Is that what you intended?

Any rerun of **Meson** CI will always run the meson CI builds.

mobin-2008: > Could we just leave the condition as it was ("pull request label contains 'meson'") and then if we want to run meson CI for a given pull request just add the label and then re-run CI?

I tested it before but it's need an another commit to work.

davmac314: > Any rerun of Meson CI will always run the meson CI builds.

But the condition is "or":

```
    if: ${{ github.run_attempt >= 2 || contains(github.event.pull_request.labels.*.name, 'meson') }}
```

If this is the 2nd run, OR the pull event contains 'meson'

mobin-2008: **Regular CI** and **Meson based CI** are two different workflows and rerun of one of them, doesn't rerun other workflow.

davmac314: Alright, so how can you re-run the meson workflow if it never ran the first time?

mobin-2008: 1. a PR is related to meson but it doesn't have meson label.
2. Meson CI skipped for first time.
3. one of the maintainers rerun skipped workflow. it's will make `github.run_attempt` = 2.
4. Meson CI will run for that PR. Also we add meson label for auto meson ci.

davmac314: Why not:

1. a PR is related to meson but it doesn't have meson label.
2. Meson CI skipped for first time.
3. one of the maintainers adds the  meson label
4. maintainer then reruns the skipped workflow

Doesn't that achieve the same thing?

mobin-2008: 5. Unfortunately it's skipped again due to some reasons.

davmac314: > > So _any_ rerun of CI will always run the meson CI builds? Is that what you intended?
> 
> Any rerun of **Meson** CI will always run the meson CI builds.

I don't see how to run just Meson CI. Github UI says "rerun all jobs". It's not specific to a workflow.

image

davmac314: So if the condition is as it is currently, re-running jobs will run the meson CI workflow even if it isn't a meson-related pull request.

davmac314: Hmm, maybe I'm wrong. Maybe it separates different workflows into different jobs. But I want to check more, will need to do that later.

mobin-2008: For sure, I reran Regular CI for your last commit. As you see It doesn't trigger Meson based CI.
https://github.com/davmac314/dinit/actions/runs/5115941484

davmac314: Ok, sounds good.

Can you please put a comment in the file explaining all this (just a couple of sentences is fine).

Can we make it run on any change to master (once merged/pushed)? I think it's good to know if meson build does break

davmac314: Details: https://blog.mergify.com/running-github-actions-only-on-certain-pull-requests/

mobin-2008: I think we need to bypass that check in workflow for pushes, No? Currently it needs a re-run request to run.

davmac314: > Currently it needs a re-run request to run.

I think with the correct dispatch (`on: pull_request: types: - labeled') it won't need a re-run request.

But yes, the conditions may need to be different for a pull request vs push to master.

mobin-2008: But seems like it doesn't work on newer commits!

davmac314: you need it dispatched on the default event types too - `opened`, `reopened` and `synchronize`. If the event type is `label` it only triggers at the time the label is added.

PR #198, opened 2023-06-08
igr-tests: Refactoring

mobin-2008:
Hi!
Dinit currently use a simple runner for running some scripts as "integration tests". but it system needs refactoring.
> Why?

Because it's have many problems and bugs [Highest priority to lowest]:
1. Tests who use dinit socket may hangs forever. look at this example:
     https://github.com/davmac314/dinit/blob/b9f20c58a23a50493d2cea07985cadb25381de5f/src/igr-tests/force-stop/run-test.sh#L5-L11
    Current igr-tests doesn't have `set -e` so if a command fails it doesn't exit immediately so it hangs forever and need manual cancellation. Of course if command runs successfully but refuse to create socket, it hangs again. I think it's a bug due to poor design.

2. Many statements don't have error/warning messages. error messages are very useful in many statements, such as when output file isn't exist, It helps developers to faster debugging.

3. Many things are repeated in tests, Things who can be in a reference file (such as launching dinit, stopping dinit and ...).

4. Current implementation isn't Meson compatible. For meson builds we just copy entire igr-tests into build environment and run tests. I think that's can be changed if we have something like dynamic `$TEMP`.

5. And many other things such as all igr-tests do same tasks via different ways without any reason (some scripts disable dinit via `kill` and some other via `dinitctl` and ...)

I think it's definitely reasonable.

Regards.

# ToDo
- [x] Create initial `igr_functions`
- [x] ~Refactor `compare_*()` functions (Or even remove them)~ Not needed.
- [x] support `DEBUG` keyword for avoid `--quiet` in dinit commands. (meson done, for Make is pending)
- [x] `set -eu` for all tests
- [x] Refactor all scripts
- [x] Enable tests for meson
  - [x] Fix `TEST_NAME` variable
- [x] Complete docs
- [x] Fix environ2 on MacOS
- [x] Fix `reload2` fails for Meson builds. (It looks weird)
- [x] Complete commit description

Changes:
1. `77` return code now used for skipping tests for better compatibility with GNU testing protocol (and Meson). Also `2` used by shell for reporting `set -u` failures.

2. All tests now use `set -eu`.

3. `$STATUS` Removed in favor of new error handling.

4. All tests now use `$TEMP/output` for temp files (such as `basic-ran`).

5. Some documented wrapper functions available in `src/igr-tests/igr_functions` to do some most repeated things (such as spawn a dinit instance, comparing a file with text & ...).


src/igr-tests/environ2/run-test.sh
```diff
@@ -1,41 +1,43 @@
 #!/bin/sh
 
-cd "$(dirname "$0")"
+# FIXME: UID (and possibly GID) cannot unsent on MacOS! Skipping...
```
mobin-2008, line 3: `$UID` is readonly variable on MacOS. We can just ignore this and don't unseting `$UID`?

src/igr-tests/meson.build
```diff
@@ -1,33 +1,31 @@
...
-    '-r',
-    './',
-    meson.current_build_dir()
-    ],
-    check : true
-)
+igr_tests_env += [
+     'IGR_FUNCTIONS=@0@/igr_functions'.format(meson.current_source_dir()),
+     'TEMP=@0@'.format(meson.current_build_dir()),
+     'DEBUG=yes'
```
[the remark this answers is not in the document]
davmac314: Need more info! Which test, what is the dinit command line?

src/igr-tests/environ2/run-test.sh
davmac314: Yes just ignore it

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,231 @@
+# Included from run-test.sh files.
+# shellcheck disable=SC2034,SC2148 # Disable some false warnings/errors.
+
+### This file contain basic variables for running Integration tests of
+### Dinit.
+
+# SPDX-License-Identifier: Apache-2.0
```
davmac314, line 7: Remove this please, we aren't using SPDX identifiers in the rest of the code base, why start doing it here?

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,231 @@
+# Included from run-test.sh files.
```
davmac314, line 1: Can we name this file `igr_functions.sh` so it is clear what it contains?

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,231 @@
...
+# Also if you need to capture stderr, set something as $3 (such as "err").
+# RESULT: set $CMD_OUT variable contain command output.
+#         return 0 if command output is the same with content of file.
+#         return 1 if command output isn't the same with content of file.
+#         throw an error on any other failure.
+compare_cmd() {
+    if [ ! -e "$2" ]; then
+        error "$2 file isn't exist!"
+    fi
+    if [ -n "${3:-}" ]; then
```
davmac314, line 221: Make this look for something specific. Eg `if [ "${3:-}" == "err" ]; then`. That gives the possibility to add other options later and also to check if an invalid option is supplied (though you don't have to do that now).

src/igr-tests/meson.build
```diff
@@ -1,33 +1,31 @@
...
-    'cp',
-    '-r',
-    './',
-    meson.current_build_dir()
-    ],
-    check : true
-)
+igr_tests_env += [
+     'IGR_FUNCTIONS=@0@/igr_functions'.format(meson.current_source_dir()),
+     'TEMP=@0@'.format(meson.current_build_dir()),
```
davmac314, line 5: Please call the output directory something more specific than TEMP. "IGR_OUTPUT" would be a good choice.

src/igr-tests/igr_functions
mobin-2008: I thought maybe SPDX will useful. I'll remove it.

src/igr-tests/meson.build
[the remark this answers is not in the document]
davmac314: Please debug issues like this before opening a PR (next time). I am working on other things and since it's your own changes causing the issue it's up to you to debug it.

You should probably check what command line is actually being passed to `dinit` in these cases as a start.

davmac314:
I will unsubscribe now, mention me if you have a question or need review

src/igr-tests/meson.build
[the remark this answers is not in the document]
davmac314: (to be clear, I don't expect you to debug `dinit` itself, but at least figure out how it is being run that makes it behave this way)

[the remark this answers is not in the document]
mobin-2008: Seems like it cause by double-quoted `$QUIET`. Fixed.

mobin-2008:
@davmac314 This PR is ready. Also Debian 12 will be released soon, Is it good to include that in Dinit CI? or even replace Debian 11 with that?

src/igr-tests/igr-runner.cc
```diff
@@ -18,9 +18,9 @@ int main(int argc, char **argv)
             "before-after"};
     constexpr int num_tests = sizeof(test_dirs) / sizeof(test_dirs[0]);
 
-    int passed = 0;
-    int skipped = 0;
-    int failed = 0;
+    unsigned short int passed = 0;
+    unsigned short int skipped = 0;
+    unsigned short int failed = 0;
```
davmac314, line 23: Please revert this. It really has no benefit. We don't need to save 6 bytes and anyway it doesn't since the executable size increases (on x86-64) by way more.

src/igr-tests/Makefile
```diff
@@ -1,16 +1,13 @@
 include ../../mconfig
 
 check-igr: igr-runner
-	DINIT_EXEC=../../dinit DINITCTL_EXEC=../../dinitctl DINITCHECK_EXEC=../../dinitcheck DINITMONITOR_EXEC=../../dinit-monitor ./igr-runner
+	IGR_FUNCTIONS="../igr_functions" EXEC_PREFIX="../../" MAKE=1 ./igr-runner
```
davmac314, line 4: Doesn't look like "MAKE=1" is needed, can we remove it?

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+# According to POSIX, echo has some unspecified behavior in some cases, for example
+# when its first argument is "-n" or backslash ("\"). We prefer to rely on
+# POSIX documented things.
+# So we replace shell built-in echo with a printf based function.
+# For more info see http://www.etalabs.net/sh_tricks.html
+echo() {
+    IFS=" " printf %s\\n "$*"
+}
+error() {
+    >&2 echo "${TEST_NAME:-}: Error: $*"
```
davmac314, line 17: This uses `"$*"` which will expand all parameters, but then just it specifically checks for a 2nd parameter (`if [ -n "${2:-}" ]; then`). So probably you want to drop the "sub message" part or change `"$*"` to `"$1"`.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+    if [ -n "${2:-}" ]; then
+        >&2 echo " ... $2"
+    fi
+    echo
+}
+
+## Executable path resolvers functions
+# These functions return 0 and set suitable variable when program found.
+# And return 1 on failure.
+__find_dinit() {
```
davmac314, line 43: use consistent naming, remove the "__" prefix

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+    return 0
+}
+
+## Change directory to run-test.sh location
+cd "$(dirname "$0")" || error "Can't change directory to script location."
+
+## Prepare $IGR_OUTPUT
+TEST_NAME="${PWD##*/}"
+if [ -z "$TEST_NAME" ]; then error "Failed to guess test name."; fi
+if [ -n "${MAKE:-}" ]; then
```
davmac314, line 90: Rather than a specific `MAKE` variable just check if `IGR_OUTPUT` is not already set. If it's not set you can assume it's a make-based build (or someone running the script directly, which I do when I am trying to solve a failure in a single test).

src/igr-tests/Makefile
davmac314: Ok I see where it is used (in igr_functions) but I think it should be removed, see other comment.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+if [ -n "${DEBUG:-}" ]; then
+    QUIET=""
+else
+    QUIET="--quiet"
+fi
+
+## Integreation tests helper functions
+# This function spawn a dinit daemon with temp socket.
+# It's pass its flags as dinit daemon flags.
+# RESULT: return 0 & sets "SOCKET" & "DINITPID" variables on sccuess.
```
davmac314, line 111: `sccuess` -> `success`

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+#         throw an error on failure.
+spawn_dinit() {
+    __find_dinit || error "Cannot find dinit exec path." "Check 'EXEC_PREFIX' variable value"
+    "$DINIT" $QUIET -u -d sd -p "$SOCKET" "$@" &
```
davmac314, line 115: Here it uses "$SOCKET" but the comment says that the function _sets_ SOCKET

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+fi
+mkdir -p "$IGR_OUTPUT"/output/
+export SOCKET="$IGR_OUTPUT/socket"
+if [ -n "${DEBUG:-}" ]; then
+    QUIET=""
+else
+    QUIET="--quiet"
+fi
+
+## Integreation tests helper functions
```
davmac314, line 108: 'Integreation' -> 'Integration'

src/igr-tests/meson.build
```diff
@@ -99,6 +104,7 @@ test(
      'reload2',
      reload2_test_script,
      env : igr_tests_env,
+     is_parallel : false,
```
davmac314, line 107: Why is this needed?

src/igr-tests/reload2/run-test.sh
```diff
@@ -1,56 +1,44 @@
 #!/bin/sh
 
-cd "$(dirname "$0")"
+set -eu
+. "$IGR_FUNCTIONS"
+
+# FIXME: Fails on meson
```
davmac314, line 6: I guess this is was meant to be removed?

davmac314 (changes requested):

src/igr-tests/igr-runner.cc
davmac314: The main issue I have is that it is not idiomatic. In C/C++ we use `int` (or sometimes `unsigned int`) when an integer is needed. `short` is normally only ever used in a `struct` for example that may have many instances and so the space saving becomes important. But for a single variable, it's not, and it can be less efficient because it may need more or longer machine instructions.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,232 @@
...
+# It's pass its flags as dinitcheck flags.
+# RESULT: Any return code from dinitcheck on sccuess.
+#         Throw an error on failure.
+dinitcheck() {
+    __find_dinitcheck || error "Cannot find dinitcheck exec path." "Check 'EXEC_PREFIX' variable value."
+    "$DINITCHECK" -d sd "$@"
+}
+
+# This function compare content of given file with expected result.
+# Accpets file as $1 and a text as $2.
```
davmac314, line 174: Accpets -> Accepts

src/igr-tests/igr-runner.cc
davmac314: Example: https://godbolt.org/z/xvf7aY7WY

When `TTT=int` it's fewer instructions than when `TTT=unsigned short`. You can lose much more memory in extra instructions than you gain by reducing `int` to `short`.

But also, as a general principle: if it isn't broken, don't fix it. There is nothing wrong with the code as it was.

davmac314:
This mostly looks good, I pointed out a few minor things and I had a question about `is_parallel` for one of the tests.

src/igr-tests/meson.build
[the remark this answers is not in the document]
davmac314: Right, `"$VAR"` expands to an empty argument (not no argument) if VAR is empty. However this does reveal a minor dinit bug: it should probably check for empty arguments and not try to start a service with no name.

davmac314:
> @davmac314 This PR is ready. Also Debian 12 will be released soon, Is it good to include that in Dinit CI? or even replace Debian 11 with that?

It's not necessary. We don't want CI to be building bleeding-edge distribution releases since any problems are more likely to be with the distro and not with dinit.

src/igr-tests/meson.build
mobin-2008: I missing this one.

src/igr-tests/before-after/run-test.sh
```diff
@@ -1,46 +1,44 @@
...
-mkdir -p output
-rm -rf output/*
+rm -rf "$IGR_OUTPUT"/output/*
 
-"$DINIT_EXEC" -q -d sd1 -u -p socket  &
+spawn_dinit -d sd1
 
-# Give some time for startup
-sleep 0.2
+# Custom error.
```
davmac314, line 12: Why is it needed?

src/igr-tests/before-after/run-test.sh
```diff
@@ -1,46 +1,44 @@
...
 # start parent; should start service2 and then service1 (due to before= in service2).
-"$DINITCTL_EXEC" --quiet -p socket start parent
+dinitctl $QUIET start parent || __error
 
-if [ "$(cat output/script-output)" != "$(printf "two\none\n")" ]; then
-    "$DINITCTL_EXEC" --quiet -p socket shutdown
-    return 1
+if ! compare_text "$IGR_OUTPUT"/output/script-output "$(printf "two\none\n")"; then
+    stop_dinit
+    error "$IGR_OUTPUT/output/script-output didn't contain expected result!"
```
davmac314, line 23: `error` already calls `stop_dinit` so why do both? It looks like if `stop-dinit` is called twice it will cause an error

src/igr-tests/before-after/run-test.sh
davmac314: (Also, why the "__" prefix on the name?)

src/igr-tests/igr_functions
davmac314: You've marked this as resolved, but I don't see the change. The comment above this function still says that it sets SOCKET, but it doesn't set SOCKET.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,233 @@
...
+    done
+    return 0
+}
+
+# This function stop current Dinit instance.
+# Doesn't accept anything.
+# RESULT: return 0 if Dinit stops.
+#         throw an error on failure.
+# NOTE: Don't use error(), using error() will result a inifinite cycle.
+stop_dinit() {
```
davmac314, line 136: If this is called when dinit has already stopped, it will error out. Could you make it handle that more gracefully? For example if the socket doesn't exist, don't try to stop dinit.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,233 @@
...
+# This function compare content of given file with expected result.
+# Accepts file as $1 and a text as $2.
+# RESULT: return 0 if content of file is the same with given text.
+#         return 1 if content of file is not the same with given text.
+#         throw an error on any other failure.
+compare_text() {
+    if [ ! -e "$1" ]; then
+        error "$1 file isn't exist!"
+    fi
+    if [ "$(cat "$1" || error "Can't read given file!")" = "$2" ]; then
```
davmac314, line 183: `$(cat "$1" || error "Can't read given file!")`

This doesn't work. If cat fails, `error` gets called inside the subshell and won't cause the original shell to exit.

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,233 @@
...
+        return 0
+    else
+        return 1
+    fi
+}
+
+# This function compare content of given file with expected result in another file.
+# Accpets file as $1 and another file as $2.
```
davmac314, line 191: Accpets -> Accepts

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,233 @@
...
+    fi
+    if cmp -s "$1" "$2"; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+# This function compare command result with content of a file.
+# Accpets command as $1 and expected file as $2.
```
davmac314, line 210: Accpets -> Accepts

src/igr-tests/basic/sd/boot
davmac314, line 1: It's not necessary to rename this file, just start the service called "basic".

Try to avoid unnecessary changes.

davmac314 (changes requested):
A few more changes.

You may also need to rebase on current master.

src/igr-tests/igr_functions
mobin-2008: Is it a good check?
```sh
# return 0 and throw a warning if Dinit already stopped. 
    if [ ! -e "$SOCKET" ]; then
        warning "stop_dinit() called but cannot find any running dinit instance!"
        return 0
    fi
```

src/igr-tests/igr_functions
davmac314: Still need to resolve this properly

src/igr-tests/igr_functions
davmac314: Yes, that's good

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,239 @@
...
+spawn_dinit_oneshot() {
+    find_dinit || error "Cannot find dinit exec path." "Check 'EXEC_PREFIX' variable value."
+    "$DINIT" $QUIET -u -d sd -p "$SOCKET" "$@"
+}
+
+# This function find dinitctl and allow access to dinit daemon via dinitctl.
+# It's pass its flags as dinitctl flags.
+# RESULT: Any return code from dinitctl on sccuess.
+#         Throw an error on failure.
+dinitctl() {
```
davmac314, line 165: Please rename this - it's very confusing that tests run "dinitctl" but they are actually calling this function instead of running the real `dinitctl`

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,239 @@
...
+    find_dinitctl || error "Cannot find dinitctl exec path." "Check 'EXEC_PREFIX' variable value."
+    "$DINITCTL" -p "$SOCKET" "$@"
+}
+
+# This function find dinitcheck and allow access to dinitcheck.
+# It's pass its flags as dinitcheck flags.
+# RESULT: Any return code from dinitcheck on sccuess.
+#         Throw an error on failure.
+dinitcheck() {
```
davmac314, line 174: This should also be renamed

src/igr-tests/igr_functions
davmac314: Maybe `run_dinitctl`

src/igr-tests/igr_functions
mobin-2008: I mention it's spawn dinit on "$SOCKET" socket path instead of setting "$SOCKET".

src/igr-tests/igr_functions
```diff
@@ -0,0 +1,239 @@
...
+spawn_dinit() {
+    find_dinit || error "Cannot find dinit exec path." "Check 'EXEC_PREFIX' variable value"
+    "$DINIT" $QUIET -u -d sd -p "$SOCKET" "$@" &
+    DINITPID=$!
+    # Wait for Dinit socket shows up.
+    TIMEOUT=0
+    while [ ! -e "$SOCKET" ]; do
+        if [ $TIMEOUT -le 600 ]; then
+            sleep 0.1
+            TIMEOUT=$((TIMEOUT+=1))
```
davmac314, line 123: `+=` isn't required to be supported by POSIX (and I think you meant to use just `+` anyway).

davmac314 (changes requested):
Looking good, just one more change needed

src/igr-tests/igr_functions
davmac314: Actually looks like I'm wrong, POSIX does require `+=`. But since you have assignment you don't need `+=` here.

mobin-2008: Done

davmac314 (approved):

PR #199, opened 2023-06-17
dinit-service: `logfile-permissions`, `logfile-uid`, `logfile-gid` opts

### LINKED ISSUE #193: Ability to configure logfile ownership and permissions
I would like to be able to configure the ownership and permissions of log files.

This is my use case:
* I run `dinit` as the root user but under `--user` mode as I'm using it just as a service manager.
* I run as root because I need to launch `nginx` as root in order to use privileged ports.
* I have a number of other service running as unprivileged users.
* I would like for an unprivileged user to be able to read the stdout logs (using `logfile` option).

It seems that all `logfile` logs are owned by the user that launches `dinit` and are only readable by that user (`0600` permissions). I think it would be more useful (to me at least) if these logs were owned by the `run-as` user.
Is there a way to change the ownership and permissions of these files through `dinit` service descriptions? Either by `run-as` or a more general option? If there isn't, can this functionality be added?

mobin-2008:
Hi!

Some new options in services: `logfile-permissions`, `logfile-uid`, `logfile-gid`.

ToDo
- [x] Options
- [x] Fix weird `assert()` failure
- [x] Commit message
- [x] Docs and ~changelog~ notes

It's mostly ready but We need to discuss about some things:
- [x] Perms, ~uid and gid~ will changed for each run, so if a logfile exist, those things will changed to service options or dinit defaults. I think it's better than old behavior (keep file perms). if it's ok I need to note this thing in commit message, changelog and 0.17 release notes.
- [x] ~New `option` for match logfile-uid with run-as user uid?~ (Not needed)

Best regards
Fixes #193

### DIFF
--- a/doc/manpages/dinit-service.5.m4
+++ b/doc/manpages/dinit-service.5.m4
@@ -360,6 +360,23 @@ This setting has no effect if the service is set to run on the console (via the 
 \fBstarts\-on\-console\fR, or \fBshares\-console\fR options).
 The value is subject to variable substitution (see \fBVARIABLE SUBSTITUTION\fR).
 .TP
+\fBlogfile\-permissions\fR = \fIoctal-permissions-mask\fR
+Gives the permissions for the logfile specified using \fBlogfile\fR. Normally this will be 600 (user access
+only), 640 (also readable by the group), or 644 (readable by all users).
+The default is 600.
+If the logfile already exists when the service starts, its permissions will be changed to those specified with this setting.
+.TP
+\fBlogfile\-uid\fR = {\fInumeric-user-id\fR | \fIusername\fR}
+Specifies the user (name or numeric ID) that should own the logfile.
+If \fBlogfile\-uid\fR is specified as a name without also specifying \fBlogfile-gid\fR, then
+the logfile group is the primary group of the specified user (as found in the
+system user database, normally \fI/etc/passwd\fR).
+If the logfile already exists when the service starts, its ownership will be changed according to the value of this setting.
+The default value is the user id of the \fBdinit\fR process.
+.TP
+\fBlogfile\-gid\fR = {\fInumeric-group-id\fR | \fIgroup-name\fR}
+Specifies the group of the logfile. See discussion of \fBlogfile\-uid\fR.
+.TP
 \fBoptions\fR = \fIoption\fR...
 Specifies various options for this service. See the \fBOPTIONS\fR section.
 This directive can be specified multiple times to set additional options.
--- a/src/includes/load-service.h
+++ b/src/includes/load-service.h
@@ -467,6 +467,24 @@ inline gid_t parse_gid_param(unsigned line_num, const std::string &param, const 
     return grent->gr_gid;
 }
 
+// Parse a permission param, specified as a octal number of permission (such as 0600)
+inline int parse_perms(unsigned line_num, string &paramval, const std::string &servicename,
+        const char * paramname)
+{
+    std::size_t ind = 0;
+    try {
+        int perms = std::stoi(paramval, &ind, 8);
+        if (ind != paramval.length()) {
+            throw std::logic_error("");
+        }
+        return perms;
+    }
+    catch (std::logic_error &exc) {
+        throw service_description_exc(servicename, std::string(paramname) + ": badly-formed or "
+                "out-of-range numeric value", line_num);
+    }
+}
+
 // Parse a time, specified as a decimal number of seconds (with optional fractional component after decimal
 // point or decimal comma).
 inline void parse_timespec(unsigned line_num, const std::string &paramval, const std::string &servicename,
@@ -813,6 +831,10 @@ class service_settings_wrapper
     list<std::string> after_svcs;
     log_type_id log_type = log_type_id::NONE;
     string logfile;
+    int logfile_perms = 0600;
+    uid_t logfile_uid = -1;
+    gid_t logfile_uid_gid = -1; // Primary group of logfile owner if known
+    gid_t logfile_gid = -1;
     unsigned max_log_buffer_sz = 4096;
     service_flags_t onstart_flags;
     int term_signal = SIGTERM;  // termination signal
@@ -958,6 +980,11 @@ class service_settings_wrapper
         // If socket_gid hasn't been explicitly set, but the socket_uid was specified as a name (and
         // we therefore recovered the primary group), use the primary group of the specified user.
         if (socket_gid == (gid_t)-1) socket_gid = socket_uid_gid;
+        // Also for logfile_uid/gid, we reset logfile ownership to dinit process uid/gid if uid/gid
+        // wasn't specified by service
+        if (logfile_uid == (uid_t) -1) logfile_uid = getuid();
+        if (logfile_uid_gid == (gid_t)-1) logfile_uid_gid = getgid();
+        if (logfile_gid == (gid_t)-1) logfile_gid = logfile_uid_gid;
         // likewise for "run as" gid/uid, but only if we aren't supporting supplementary group initialisation
         // (if we do support supplementary groups, run_as_gid==-1 means "use the user groups including
         // supplementary groups" whereas run_as_gid==X means "use group X with no supplementary groups").
@@ -1020,17 +1047,7 @@ void process_service_line(settings_wrapper &settings, const char *name, string &
     }
     else if (setting == "socket-permissions") {
         string sock_perm_str = read_setting_value(line_num, i, end, nullptr);
-        std::size_t ind = 0;
-        try {
-            settings.socket_perms = std::stoi(sock_perm_str, &ind, 8);
-            if (ind != sock_perm_str.length()) {
-                throw std::logic_error("");
-            }
-        }
-        catch (std::logic_error &exc) {
-            throw service_description_exc(name, "socket-permissions: badly-formed or "
-                    "out-of-range numeric value", line_num);
-        }
+        settings.socket_perms = parse_perms(line_num, sock_perm_str, name, "socket-permissions");
     }
     else if (setting == "socket-uid") {
         string sock_uid_s = read_setting_value(line_num, i, end, nullptr);
@@ -1075,6 +1092,18 @@ void process_service_line(settings_wrapper &settings, const char *name, string &
         if (!settings.logfile.empty() && settings.log_type == log_type_id::NONE) {
             settings.log_type = log_type_id::LOGFILE;
         }
+    }
+    else if (setting == "logfile-permissions") {
+        string log_perm_str = read_setting_value(line_num, i, end, nullptr);
+        settings.logfile_perms = parse_perms(line_num, log_perm_str, name, "logfile-permissions");
+    }
+    else if (setting == "logfile-uid") {
+        string log_uid_s = read_setting_value(line_num, i, end, nullptr);
+        settings.logfile_uid = parse_uid_param(line_num, log_uid_s, name, "logfile-uid", &settings.logfile_uid_gid);
+    }
+    else if (setting == "logfile-gid") {
+        string log_gid_s = read_setting_value(line_num, i, end, nullptr);
+        settings.logfile_gid = parse_gid_param(line_num, log_gid_s, name, "logfile-gid");
     }
     else if (setting == "log-type") {
         string log_type_str = read_setting_value(line_num, i, end);
--- a/src/includes/proc-service.h
+++ b/src/includes/proc-service.h
@@ -176,7 +176,10 @@ class base_process_service : public service_record
     string env_file;          // file with environment settings for this service
 
     log_type_id log_type = log_type_id::NONE;
-    string logfile;           // log file name, empty string specifies /dev/null
+    string logfile;          // log file name, empty string specifies /dev/null
+    int logfile_perms = 0;   // logfile permissions("mode")
+    uid_t logfile_uid = -1;  // logfile owner user id
+    gid_t logfile_gid = -1;  // logfile group id
     unsigned log_buf_max = 0; // log buffer maximum size
     unsigned log_buf_size = 0; // log buffer current size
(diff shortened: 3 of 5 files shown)


src/baseproc-service.cc
```diff
@@ -111,6 +111,26 @@ bool base_process_service::start_ps_process(const std::vector<const char *> &cmd
...
+        int temp_fd = bp_sys::open(logfile, O_CREAT);
+        if (temp_fd == 1) {
+            log(loglevel_t::ERROR, get_name(), ": Error opening temp logfile file descriptor: ",
+                strerror(errno));
+            return false;
+        }
+        if (fchown(temp_fd, logfile_uid, logfile_gid)) {
+            log(loglevel_t::ERROR, get_name(), ": Error setting logfile owner/group: ",
+                strerror(errno));
+            close(temp_fd);
```
mobin-2008, line 124: Seems like I forgot to use `bp_sys::close()`

Duncaen:
You have to be somewhat careful about following symlinks or use `O_CREAT|O_EXCL`. A user might point the log file to a directory some user can write to and then dinit might follow a symlink to a directory the user can't write to, create the log file in it and then make it user writable.

davmac314:
> Perms, uid and gid will changed for each run, so if a logfile exist, those things will changed to service options or dinit defaults. I think it's better than old behavior (keep file perms). if it's ok I need to note this thing in commit message, changelog and 0.17 release notes.

Ok, go with this new behaviour.

> New option for match logfile-uid with run-as user uid?

Now I think about it, it probably would not be very useful. So no, don't bother with this.

NEWS
```diff
@@ -5,6 +5,9 @@ Current sponsors include github users: brentfrow, brazeon, and q66 (Daniel Koles
 Chimera Linux). Development of Dinit has also received financial support from Artix Linux.
 I am grateful for all the support, both financial and otherwise, that has been received.
 
+NOTE: This release has some minor backwards-incompatible changes about logfiles. Please read
```
davmac314, line 8: This note isn't needed. There are several backwards-incompatible changes, not all of them to do with logfiles, and they are already listed in the changes section. (I will revise these notes before release. You don't need to add changes to the NEWS as part of a pull request).

src/run-child-proc.cc
```diff
@@ -211,6 +211,10 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
             if (output_fd == -1) {
                 output_fd = open(logfile, O_WRONLY | O_CREAT | O_APPEND, S_IRUSR | S_IWUSR);
                 if (output_fd == -1) goto failure_out;
+                if (this->log_type == log_type_id::LOGFILE) {
```
davmac314, line 214: It was difficult for me to figure out why this check is necessary. I think it needs a comment ("if log type is NONE, we don't want to change ownership/permissions of /dev/null" or something similar).

src/includes/service.h
```diff
@@ -314,6 +314,12 @@ class service_record
     uid_t socket_uid = -1;  // socket user id or -1
     gid_t socket_gid = -1;  // socket group id or -1
 
+    string logfile;      // path to the logfile for record service outputs
+    int logfile_perms = 0;      // logfile permissions("mode")
+    uid_t logfile_uid = -1;      // logfile user id or -1
+    uid_t logfile_uid_gid = -1;  // logfile primary group id or -1
```
davmac314, line 320: This (`logfile_uid_gid`) shouldn't be here - it is never set, never used.

davmac314 (changes requested):

src/run-child-proc.cc
```diff
@@ -211,6 +211,10 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
...
+                    if (fchown(output_fd, logfile_uid, logfile_gid) == -1 ) goto failure_out;
```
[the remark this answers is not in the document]
mobin-2008: Something like this?
```c++
uid_t logfile_uid = getuid();
uid_t logfile_uid_gid = getgid();
gid_t logfile_gid = getgid();
```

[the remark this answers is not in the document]
davmac314: `getuid()` and `getgid()` are the right functions to use, if you mean in `service_settings_wrapper`, yes. But don't call getgid() twice, just call it once and use the value from the first time to assign the 2nd variable.

src/includes/service.h
davmac314: Also, none of these settings belongs in `service_record`. Logging is specific to process-based services, so these belong in `base_process_service`. (Adding `logfile` here duplicates the variable that is already defined there).

mobin-2008: So I need to remove code about these things in `run-child-proc.cc`? I think yes

davmac314: No! You need to move these variables to the `base_process_service` class (except logfile, which is already there, and logfile_uid_gid, which isn't needed at all).

src/includes/load-service.h
```diff
@@ -957,7 +978,9 @@ class service_settings_wrapper
 
         // If socket_gid hasn't been explicitly set, but the socket_uid was specified as a name (and
         // we therefore recovered the primary group), use the primary group of the specified user.
+        // Same for logfile_uid/logfile_gid
         if (socket_gid == (gid_t)-1) socket_gid = socket_uid_gid;
+        if (logfile_gid == (gid_t)getgid()) logfile_gid = logfile_uid_gid;
```
mobin-2008, line 983: I think most issues are resolved but one remaining at least: If `logfile-uid` has been set, use primary group of that user as `logfile-gid` otherwise use value of `logfile-gid` or dinit process gid. In current implementation, logfile-gid completely ignored if `logfile-uid` has been set.  Hmm looks weird to fix.

src/includes/load-service.h
```diff
@@ -467,6 +467,23 @@ inline gid_t parse_gid_param(unsigned line_num, const std::string &param, const
     return grent->gr_gid;
 }
 
+// Parse a permission param, specified as a decimal number of permission (such as 0600)
```
davmac314, line 470: The permissions are specified as an octal, not a decimal, the comment is wrong

src/includes/load-service.h
davmac314: Why not use the same logic as for socket? -1 can be the special value to mean "not (yet) set". You can initialise `logfile_uid` and `logfile_gid` to -1, set `logfile_gid` to `logfile_uid_gid` if it's still -1 here, and set it to `getgid()` if it's still -1 after that (as well as setting `logfile_uid` to `getuid()` if it is -1).

In general, please try to use similar logic to other parts of the code when it makes sense. We shouldn't do one thing one way and a slightly different thing a completely different way, if they can be done the same way, unless there is a good reason.

src/includes/load-service.h
```diff
@@ -1020,17 +1043,7 @@ void process_service_line(settings_wrapper &settings, const char *name, string &
...
-            settings.socket_perms = std::stoi(sock_perm_str, &ind, 8);
-            if (ind != sock_perm_str.length()) {
-                throw std::logic_error("");
-            }
-        }
-        catch (std::logic_error &exc) {
-            throw service_description_exc(name, "socket-permissions: badly-formed or "
-                    "out-of-range numeric value", line_num);
-        }
+        settings.socket_perms = parse_perms(line_num, sock_perm_str, name, "socket_permissions");
```
davmac314, line 1046: socket_permissions should be socket-permissions

src/includes/proc-service.h
```diff
@@ -177,6 +177,9 @@ class base_process_service : public service_record
 
     log_type_id log_type = log_type_id::NONE;
     string logfile;           // log file name, empty string specifies /dev/null
+    int logfile_perms = 0;         // logfile permissions("mode")
+    uid_t logfile_uid = getuid();  // logfile user id or dinit daemon uid
```
davmac314, line 181: Don't initialise these here via getuid() etc, that's wasteful, they will be set via `set_logfile_details` when it is called. You can leave them as -1 here.

The comment should describe the purpose of the variable. It currently says "logfile user id _or_ dinit daemon uid" but the variable is _always_ for the logfile uid, the daemon uid is just the default value.

src/includes/service.h
davmac314: `logfile` is still here

src/load-service.cc
```diff
@@ -648,7 +648,8 @@ service_record * dirload_service_set::load_reload_service(const char *name, serv
             rvalps->set_run_as_uid_gid(settings.run_as_uid, settings.run_as_gid);
             rvalps->set_notification_fd(settings.readiness_fd);
             rvalps->set_notification_var(std::move(settings.readiness_var));
-            rvalps->set_log_file(std::move(settings.logfile));
+            rvalps->set_logfile_details(std::move(settings.logfile), settings.logfile_perms,
+                settings.logfile_uid, settings.logfile_gid);
```
davmac314, line 652: As per CODE-STYLE:

> When wrapping a line, the indent should be increased by at least two levels on the subsequent lines

src/load-service.cc
```diff
@@ -689,7 +690,8 @@ service_record * dirload_service_set::load_reload_service(const char *name, serv
             rvalps->set_start_timeout(settings.start_timeout);
             rvalps->set_extra_termination_signal(settings.term_signal);
             rvalps->set_run_as_uid_gid(settings.run_as_uid, settings.run_as_gid);
-            rvalps->set_log_file(std::move(settings.logfile));
+            rvalps->set_logfile_details(std::move(settings.logfile), settings.logfile_perms,
+                settings.logfile_uid, settings.logfile_gid);
```
davmac314, line 694: Indentation here

src/load-service.cc
```diff
@@ -724,7 +726,8 @@ service_record * dirload_service_set::load_reload_service(const char *name, serv
...
```
davmac314, line 730: Indentation here too

src/run-child-proc.cc
```diff
@@ -211,6 +211,12 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
             if (output_fd == -1) {
                 output_fd = open(logfile, O_WRONLY | O_CREAT | O_APPEND, S_IRUSR | S_IWUSR);
                 if (output_fd == -1) goto failure_out;
+                // Set permission of logfile if present
+                // if log type is NONE, we don't want to change ownership/permissions of /dev/null!
+                if (this->log_type == log_type_id::LOGFILE) {
+                    if (fchown(output_fd, logfile_uid, logfile_gid) == -1 ) goto failure_out;
```
davmac314, line 217: There's an extra space between `-1` and `)` here, and on the next line

doc/manpages/dinit-service.5.m4
```diff
@@ -360,6 +360,23 @@ This setting has no effect if the service is set to run on the console (via the
 \fBstarts\-on\-console\fR, or \fBshares\-console\fR options).
 The value is subject to variable substitution (see \fBVARIABLE SUBSTITUTION\fR).
 .TP
+\fBlogfile\-permissions\fR = \fIoctal-permissions-mask\fR
+Gives the permissions for the logfile specified using \fBlogfile\fR. Normally this will be 600 (user access
```
davmac314, line 364: These values don't seem normal for a logfile. Wouldn't it be more likely that read-only permission is granted to the group and/or all users?

doc/manpages/dinit-service.5.m4
```diff
@@ -360,6 +360,23 @@ This setting has no effect if the service is set to run on the console (via the
...
+only), 660 (user and group access), or 666 (all users).
+The default is 600.
+If logfile exist, thats perms will be ignored and the logfile permissions will be replaced with default or this option value.
```
davmac314, line 367: The grammar on this line needs a little work and "perms" is an abbreviation, how about:

"If the logfile already exists when the service starts, its permissions will be changed to those specified with this setting."

doc/manpages/dinit-service.5.m4
```diff
@@ -360,6 +360,23 @@ This setting has no effect if the service is set to run on the console (via the
...
+.TP
+\fBlogfile\-uid\fR = {\fInumeric-user-id\fR | \fIusername\fR}
+Specifies the user (name or numeric ID) that should own the logfile.
+If \fBlogfile\-uid\fR is specified as a name without also specifying \fBlogfile-gid\fR, then
+the logfile group is the primary group of the specified user (as found in the
+system user database, normally \fI/etc/passwd\fR).
+If the \fBlogfile\-uid\fR setting is not provided, the logfile will be owned by the user id of the \fBdinit\fR process.
+Also if logfile exist, thats uid will be ignored and the logfile will be owned by the user of \fBdinit\fR process or this option value.
```
davmac314, line 375: Replace lines 374-375 with:

"If the logfile already exists when the service starts, its ownership will be changed according to the value of this setting.
The default value is the user id of the \fBdinit\fR process."

src/includes/load-service.h
```diff
@@ -467,6 +467,23 @@ inline gid_t parse_gid_param(unsigned line_num, const std::string &param, const
...
+        if (ind != paramval.length()) {
+            throw std::logic_error("");
+        }
+        return perms;
+    }
+    catch (std::logic_error &exc) {
+        throw service_description_exc(servicename, std::string(paramname) + ": badly-formed or "
+                "out-of-range numeric value", line_num);
+    }
+}
```
davmac314, line 486: There should be a blank line between this method and the next one (as per style guide).

davmac314 (changes requested):

doc/manpages/dinit-service.5.m4
davmac314: For example:
"Normally this will be 600 (user access only), 640 (also readable by the group), or 644 (readable by all users)."

src/includes/service.h
mobin-2008: I saw that and removed that from `service.h`, this PR doesn't modify something in `service.h` now.

PR #208, opened 2023-07-22
control: Limit memory use by control connections.

mobin-2008:
Hi.

I working on this PR for resolving one of the last blockers for 0.17 release:
https://github.com/davmac314/dinit/blob/7d491be0778a3dbb50ba92f2c79b87d6999d978a/TODO#L3-L5

`Signed-off-by: Mobin "Hojjat' Aydinfar <mobin@mobintestserver.ir>`


davmac314:
The idea is not to reject commands if the buffer is too full, but to stop receiving them until the buffer has had a chance to reduce. That would be done by disabling the read watch (IN_EVENTS) in the `control_conn_watcher` (and then enabling the read watch again once the buffer was below a threshold).

This solution as written doesn't actually address the problem that the TODO note is concerned with, which is that a malicious or badly-written client could keep stuffing the output buffer causing it to consume an excessive amount of memory. I.e. this solution requires that the buffer management is performed by the client itself, which doesn't help.

(Note that in practice a malicious client already needs to be root to be able to send commands via the control socket, so this is mostly a theoretical concern).

Just FYI there's also an error in your implementation that means it always returns `DINIT_RP_IDLE`. This line:
```
        if (sizeof(outbuf) < 1000) {
```
... is checking the size of the `outbuf` object itself, i.e. the size of the type `list<vector<char>>`, which is a compile-time constant (that happens to be less than 1000). It does not check the number of bytes in the buffer. A proper fix will keep track of the number of bytes currently in the buffer in a member variable. This would mean updating that variable whenever something is added to or removed from the buffer. (It would be possible to calculate the size instead of tracking it in a variable, but since that requires walking through each item in the top-level `list` it is not efficient).

davmac314:
Btw. If you'd like to implementing as I've suggested please feel free to re-open this PR or create a new one - I'm just closing it for now because it's definitely not the right fix at the moment.

mobin-2008:
I updated it to disable `IN_EVENTS` when outbuf vector have >= 1000 members (Is it efficient?) and enable it again when outbuf have < 1000 members. Is it fit our requirements?

davmac314:
> I updated it to disable `IN_EVENTS` when outbuf vector have >= 1000 members (Is it efficient?) and enable it again when outbuf have < 1000 members. Is it fit our requirements?

`outbuf` is a list of vectors. See the definition:
```
    // Buffer for outgoing packets. Each outgoing back is represented as a vector<char>.
    list<vector<char>> outbuf;
```

Checking its size with `outbuf.size()` just tells you how many vectors are in the list. It does not tell you how much data is in each of the vectors.

Also, as I mentioned earlier:

> A proper fix will keep track of the number of bytes currently in the buffer in a member variable. This would mean updating that variable whenever something is added to or removed from the buffer. (It would be possible to calculate the size instead of tracking it in a variable, but since that requires walking through each item in the top-level list it is not efficient).

src/control.cc
```diff
@@ -1135,7 +1135,8 @@ void control_conn_t::service_event(service_record *service, service_event_t even
 
 bool control_conn_t::queue_packet(const char *pkt, unsigned size) noexcept
 {
-    int in_flag = bad_conn_close ? 0 : IN_EVENTS;
+    // Close connection on bad con or temporarily Limit input if necessary
+    int in_flag = bad_conn_close || (outbuf_size >= 1000) ? 0 : IN_EVENTS;
```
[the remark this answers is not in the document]
davmac314: You can probably just remove line 1158 `iob.set_watches(in_flag);`, it shouldn't be needed (I think). Then you don't need the correct value of `in_flag` there.

src/includes/control.h
```diff
@@ -111,6 +111,9 @@ class control_conn_t : private service_listener
     
     // Buffer for outgoing packets. Each outgoing back is represented as a vector<char>.
     list<vector<char>> outbuf;
+    // Output buffer size. functions will increase this value when pushing something into
+    // outbuf and lowering this value when something get removed from outbuf.
```
davmac314, line 115: I think a better comment would be "Current output buffer size in bytes", for example. You don't need to say that "functions will increase this value" etc because that is already clear. Comments should describe the purpose of the variable or function and only give further detail if it is really needed.

src/control.cc
```diff
@@ -1190,7 +1192,8 @@ bool control_conn_t::queue_packet(const char *pkt, unsigned size) noexcept
 // make them extraordinary difficult to combine into a single method.
 bool control_conn_t::queue_packet(std::vector<char> &&pkt) noexcept
 {
-    int in_flag = bad_conn_close ? 0 : IN_EVENTS;
+    // Close connection on bad con or temporarily Limit input if necessary
+    int in_flag = bad_conn_close || (outbuf_size >= 1000) ? 0 : IN_EVENTS;
```
davmac314, line 1196: Same concern here as with the other `queue_packet` function.

davmac314 (changes requested):
This is on the right track.

Can you please also use a named constant (a `constexpr int` at the top of the file is fine) for the buffer watermark value rather than using a literal `1000`. Also, 1000 is way too low. Maybe 16kb (16*1024) is reasonable.

src/control.cc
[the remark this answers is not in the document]
mobin-2008: > You can probably just remove line 1158 `iob.set_watches(in_flag);`, it shouldn't be needed (I think). Then you don't need the correct value of `in_flag` there.

I think it's needed, igr-tests failed due this at least. Seems like we need to double check `outbuf_size`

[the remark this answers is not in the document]
davmac314: Ok, that is a bug then and needs to be fixed. I may be able to fix it now.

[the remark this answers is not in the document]
mobin-2008: After merging this PR?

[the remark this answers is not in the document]
davmac314: Done, you will need to rebase. Setting `IN_EVENTS` is now only done in `control_conn_cb` in control.h, so you'll need to adjust for that. (You had missed a spot in `control_conn_t::send_data` where `IN_EVENTS` was being set, but that's gone now anyway).

davmac314 (changes requested):

src/control.cc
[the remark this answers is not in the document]
davmac314: > After merging this PR?

No, sorry, the bug fix takes precedence. It has been done now.

src/includes/control.h
```diff
@@ -250,7 +252,7 @@ inline dasynq::rearm control_conn_cb(eventloop_t * loop, control_conn_watcher *
     }
     
     int watch_flags = 0;
-    if (!conn->bad_conn_close) {
+    if (!conn->bad_conn_close || conn->outbuf_size < 16384) {
```
davmac314, line 255: The condition is wrong, it should be `&&` not `||`.

Please use a named constant instead of a literal value.

davmac314 (changes requested):

src/includes/control.h
mobin-2008: Done but I can't squash it now for some reasons, I will squash these 3 commits in merging process.

src/includes/control.h
```diff
@@ -249,8 +251,9 @@ inline dasynq::rearm control_conn_cb(eventloop_t * loop, control_conn_watcher *
         }
     }
     
+    constexpr int outbuf_limit = 16384; // 16KB
```
davmac314, line 254: > Can you please also use a named constant (a constexpr int at the top of the file is fine) 

Please place near the top of the file as I asked originally and use a name in all caps so it is clear that it is a constant (just like `IN_EVENTS` etc).

src/includes/control.h
davmac314: How are you trying to squash? What is the problem?

src/includes/control.h
davmac314: (The point of making it a named constant is so that it is easy to find and modify and it can be used elsewhere in the code if needed. If you put it right next to the code that uses it these benefits are lost).

src/includes/control.h
mobin-2008: I used my phone to fix final things. I squashed them

davmac314 (approved):

mobin-2008:
and It's done! I'm very excited for new 0.17 release and new episode of "Escape from System D" :)

davmac314:
There are a few other changes to go in yet! But yes it is one step closer. :rocket:

PR #209, opened 2023-08-02
configure: Some minor fixes and improvements

mobin-2008:
Hello. Changes:

1. `sub_warning()` is combined with `warning()` for better usage
2. Some new short options (`-h`, `-q`, `-c`)
3. Fixes "TEST_CXXFLAGS_EXTRA" which got ignored
4. Fixes unnecessary forking for writing into mconfig
5. `--clean` renamed to `--clear` (I think it's better now)

### DIFF
--- a/configure
+++ b/configure
@@ -39,16 +39,15 @@ sub_error()
     exit 1
 }
 
+# warning() have a built-in extra information feild which is optional
+# $1: Info, $2: Extra info (Optional)
+# NOTE: Always quote your warning() messages
 warning()
 {
-    echo
-    echo "Warning: $*"
-}
-
-sub_warning()
-{
-    echo "  ... $*"
-    echo
+    >&2 echo
+    >&2 echo "Warning: $1"
+    [ -n "${2:-}" ] && >&2 echo "  ... $2"
+    >&2 echo
 }
 
 cxx_works()
@@ -115,9 +114,9 @@ Generates build configuration for Dinit.
 
 Defaults for the options are specified in brackets.
 
-  --help                        This help message.
-  --quiet                       Don't print normal messages, just errors.
-  --clean                       Clear mconfig and configure's temporary files.
+  -h, --help                    This help message.
+  -q, --quiet                   Don't print normal messages, just errors.
+  -c, --clear                   Clear mconfig and configure's temporary files.
 
 Target options:
   --platform=PLATFORM           Set the platform manually (Just for cross-platform cross-compile!) [autodetected]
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -39,16 +39,13 @@ sub_error()
     exit 1
 }
 
+# NOTE: Always quote your warning() messages
 warning()
```
davmac314, line 43: Add a comment explaining what parameters can/must be supplied (i.e. warning message and optional extra information)

configure
```diff
@@ -39,16 +39,13 @@ sub_error()
...
-    echo "Warning: $*"
-}
-
-sub_warning()
-{
-    echo "  ... $*"
-    echo
+    >&2 echo
+    >&2 echo "Warning: $1"
+    [ -n "${2:-}" ] && >&2 echo " ... $2" # Built-in sub_warning()
```
davmac314, line 47: The "built-in sub_warning()" comment doesn't make sense to someone reading the code, because sub_warning doesn't exist. You could probably remove the comment altogether (once you've documented the parameters, see comment above).

configure
```diff
@@ -39,16 +39,13 @@ sub_error()
...
+    >&2 echo
 }
```
davmac314, line 49: Why combine `warning` with `sub_warning` but not `error` with `sub_error`? This is inconsistent.

configure
```diff
@@ -228,7 +224,7 @@ for arg in "$@"; do
         --disable-auto-restart|--enable-auto-restart=no) DEFAULT_AUTO_RESTART=false ;;
         CXX=*|CXX_FOR_BUILD=*|CXXFLAGS_FOR_BUILD=*|CPPFLAGS_FOR_BUILD=*\
         |LDFLAGS_FOR_BUILD=*|CXXFLAGS=*|CXXFLAGS_EXTRA=*|TEST_CXXFLAGS=*\
-        |TEST_CXXFLAGS_EXTRA|LDFLAGS=*|LDFLAGS_EXTRA=*|TEST_LDFLAGS=*\
+        |TEST_CXXFLAGS_EXTRA=*|LDFLAGS=*|LDFLAGS_EXTRA=*|TEST_LDFLAGS=*\
```
davmac314, line 227: Please separate fixes into separate commits from other refactoring/rework

davmac314 (changes requested):

configure
mobin-2008: Mainly because you can use sub_error() with info() without error() but warning() (and sub_warning()) didn't used like this.

davmac314: This seems borderline to me, but ok.

davmac314 (approved):

configure
```diff
@@ -391,11 +388,11 @@ if [ -n "${USE_UTMPX:-}" ]; then
     echo "USE_UTMPX=$USE_UTMPX" >> mconfig
 fi
 if [ -n "${CXX_FOR_BUILD:-}" ]; then
-    {
+    (
```
davmac314, line 391: This is wrong, it spawns a new subshell. The original code was correct.

PR #212, opened 2023-08-10
meson: `log-via-pipe` and `catlog` igr-tests

mobin-2008:
Add those new tests.


davmac314 (approved):
I would still prefer to run tests via `igr-runner`. As we get closer to 1.0 I plan to re-work the integration tests significantly and updating the meson build is going to be difficult to sustain.

mobin-2008:
> I would still prefer to run tests via `igr-runner`. As we get closer to 1.0 I plan to re-work the integration tests significantly and updating the meson build is going to be difficult to sustain.

Let's see

PR #213, opened 2023-08-11
dinit: show in `--version` if dinit was compiled with initgroups

mobin-2008:
I think it's useful to allow checking it via `dinit -v`

### DIFF
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -1116,6 +1116,9 @@ static void printVersion()
 #ifdef USE_UTMPX
             +1
 #endif
+#if USE_INITGROUPS
+            +1
+#endif
             ;
     if (feature_count != 0) {
         std::cout << "Supported features:"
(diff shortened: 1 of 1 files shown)


davmac314 (changes requested):
> I think it's useful to allow checking it via dinit -v

I disagree.

"intgroups" is just a function. It is not a feature. It will mean nothing to most users. If you want to list a feature based on `USE_INITGROUPS` being set, use an appropriate name for the feature.

The `USE_INITGROUPS` build option actually gates the use of two non-POSIX functions: `initgroups` and `setgroups`. But the _feature_ that these implement support for is _supplemental groups_. (We should, really, rename USE_INITGROUPS to something more appropriate.)

The commit message is wrong, there is no `-v` option. It should be `--version`.

davmac314:
If you can change the feature name to something suitable - maybe "supplemental-groups" - I'd like to get this for 0.17. We are almost there!

davmac314 (approved):

PR #216, opened 2023-08-14
CI: New ci which focus on security hardening

mobin-2008:

Also it tests configure script which is good I think.

Regards


mobin-2008:
Seems like it needs more investigation. I'll figure it out.

davmac314:
Also, please don't add yet another CI, just modify an existing one. FreeBSD would be a good choice since it's already using clang.

davmac314 (changes requested):
When I said don't add another CI, I meant don't add another task.  There are too many already IMO. We don't need three different tasks for FreeBSD, one is enough.

davmac314 (approved):

PR #220, opened 2023-08-17
igr: environ2: Ignore `unset` errors for `UID`, `GID`

mobin-2008:

### DIFF
--- a/src/igr-tests/environ2/run-test.sh
+++ b/src/igr-tests/environ2/run-test.sh
@@ -14,11 +14,11 @@ RGID=$(id -g)
 for var in USER LOGNAME SHELL; do
     unset $var
 done
-# $UID & $GID are readonly on MacOS
-if [ "$(uname)" != "Darwin" ]; then
-    unset UID
-    unset GID
-fi
+# ignore $UID & $GID unsetting errors
+# Some shells make them readonly
+for var in UID GID; do
+    unset $var > /dev/null 2>&1 || :
+done
 
 # test whether vars from global environment propagate
 export TEST_VAR="helloworld"


src/igr-tests/environ2/run-test.sh
```diff
@@ -14,11 +14,11 @@ RGID=$(id -g)
 for var in USER LOGNAME SHELL; do
     unset $var
 done
-# $UID & $GID are readonly on MacOS
-if [ "$(uname)" != "Darwin" ]; then
-    unset UID
-    unset GID
-fi
+# ignore $UID & $GID unsetting errors
+# Some systems make them readonly (such as MacOS, Void linux)
```
davmac314, line 18: It's nothing to do with Void linux, it's caused by the shell that the tests are run with (eg. bash), please adjust the comment appropriately

davmac314 (changes requested):

src/igr-tests/environ2/run-test.sh
```diff
@@ -14,11 +14,11 @@ RGID=$(id -g)
 for var in USER LOGNAME SHELL; do
     unset $var
 done
-# $UID & $GID are readonly on MacOS
-if [ "$(uname)" != "Darwin" ]; then
-    unset UID
-    unset GID
-fi
+# ignore $UID & $GID unsetting errors
+# Some shells make them readonly
```
davmac314, line 18: It would've been good to be specific here. Eg. "some shells - eg bash - make them readonly".

davmac314 (approved):

src/igr-tests/environ2/run-test.sh
mobin-2008: It actually based on system configuration for example I easily unset it on bash:
```bash
> echo $GID
1000
> unset GID
> echo $GID

>
```

mobin-2008: (or maybe bash have different behavior in non-interactive mode?)

davmac314: UID is readonly in bash. Bash doesn't set GID at all.
On MacOS, zsh sets UID and GID readonly.

PR #226, opened 2023-08-26
load-service: throw an error if the "consumer" service is its own consumer

mobin-2008:
Hi.
In the past, the dinitctl command to start that service would just hang and dinit would mark that service as STOPPED without any reason in its reports.

### DIFF
--- a/src/includes/load-service.h
+++ b/src/includes/load-service.h
@@ -1171,6 +1171,10 @@ void process_service_line(settings_wrapper &settings, const char *name, string &
     }
     else if (setting == "consumer-of") {
         string consumed_svc_name = read_setting_value(line_num, i, end);
+        if (consumed_svc_name == name) {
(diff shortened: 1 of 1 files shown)


src/includes/load-service.h
```diff
@@ -1171,6 +1171,10 @@ void process_service_line(settings_wrapper &settings, const char *name, string &
     }
     else if (setting == "consumer-of") {
         string consumed_svc_name = read_setting_value(line_num, i, end);
+        if (consumed_svc_name == name) {
+            throw service_description_exc(name, "\"consumer\" service cannot be its own consumer",
```
davmac314, line 1175: Could you change the message to just: `service cannot be its own consumer`? Otherwise, this looks good.

davmac314 (changes requested):

davmac314 (approved):

PR #229, opened 2023-08-27
configure: `--disable-strip` option

mobin-2008:
This option used by many package build systems for ensure they can create debug related packages.


davmac314:
Please don't use markdown formatting in commit messages, in future. They are not a general feature and don't work outside github. Use regular double quotes "like this" around option names etc.

davmac314 (approved):

PR #237, opened 2023-09-10
configure: fix line

Roze061:


configure
```diff
@@ -412,8 +412,8 @@ if [ "$AUTO_TEST_LDFLAGS" = "true" ] && [ "$AUTO_TEST_CXXFLAGS" = "true" ]; then
                 LDFLAGS="$established_TEST_LDFLAGS" LDFLAGS_EXTRA="$TEST_LDFLAGS_EXTRA" \
                 try_cxx_argument TEST_CXXFLAGS -fsanitize=address,undefined
     else
-        CXXFLAGS=$established_TEST_CXXFLAGS CXXFLAGS_EXTRA=$TEST_CXXFLAGS_EXTRA \
-                LDFLAGS=$established_TEST_LDFLAGS LDFLAGS_EXTRA=$TEST_LDFLAGS_EXTRA \   
+        CXXFLAGS="$established_TEST_CXXFLAGS" CXXFLAGS_EXTRA="$TEST_CXXFLAGS_EXTRA" \
```
davmac314, line 415: Can you explain why the quotes are necessary? As far as I understand, in this context no word splitting or substitution occurs on the values assigned, and that seems to be in accord with the results of a quick test.

Roze061: about quotes sorry I don't know, I saw there was a space after the slash so I deleted the space and added a quote (I actually copied and pasted).
I ran configure and there was a problem on line number 415, after I removed the space I was able to finish configure.

davmac314: You're right, the spaces shouldn't be there. I didn't see that. The quotes aren't necessary but they are in the lines above, so it's ok to add them and make it consistent. Thanks.

davmac314 (approved):

davmac314:
Merged manually- thanks for the PR!

Issue #241, opened 2023-09-12
loading service env-files relatively to the service file


q66:
The current behavior appears to treat relative paths as relative to the current working directory dinit was spawned in, which is not very useful, and also confusing.

q66:
@davmac314 what do you think about defaulting the working directory for the service also to the service dir? that would allow consistent behavior for relative commands (which means one could have e.g. a `command = ./scripts/foo.sh ...` for instance)

the current behavior is pretty undocumented, so i don't think as it is users can expect to be able to rely on anything

davmac314:
Yeah, I guess it's ok. The `command =` case is reasonably compelling. I partly think it would be better to resolve the`command =` separately though (so you can still set the working dir to something else). Anyway, what you've suggested is fine for now.

PR #242, opened 2023-09-12
Load env-files relatively to current service if not absolute path

q66:
Fixes https://github.com/davmac314/dinit/issues/241


q66:
I'm not sure if this isn't too hard of a breakage, but then I also don't see anybody using relative env-files with dinit, as they are practically useless (i guess with a system manager they're relative to root? and with a user manager, they're relative to wherever dinit was spawned, which is coincidentally also pretty useless, since it's arbitrary)

on the other hand, IMO treating these as relative to service file is intuitive and logically makes sense; there is no launch-time ambiguity

q66:
I also wonder if not to default the working dir of the service to the service file path.

PR #244, opened 2023-09-12
Default to service dir as working directory for services

q66:
@davmac314 any idea about the failing integration tests? i'm not quite seeing it


q66:
ps-environ was a forgotten thing on my side, the CI log has shown that... the other one is still failing

q66:
looks like the issue was $DINITCTL not being absolute, so i wrapped it in `realpath`

q66:
looks like macos does not have realpath, i just worked around the issue by cd'ing back from the script...

davmac314 (approved):

PR #246, opened 2023-09-16
Replace use of iostream-based classes

### LINKED ISSUE #240: Replace use of iostream-based classes
Unfortunately, C++ iostream facilities suck. They are heavyweight due to supporting different character encodings. They make it impossible to get useful error messages in a standardised way (the language spec allows for it, but the implementations apparently don't bother and map all error conditions to the same useless message). `std::ios::failure` extends `std::system_error` but doesn't use `errno` as the error code.

We should replace them (including use of `cout`/`cerr`) with a wrapper for POSIX file I/O.

mobin-2008:
Fixes #240 

Hi.
We already discussed about some technical points about this new implementation in #240 and here's its PR

## My plan
My plan has 4 sections:
low-level header: `dinit-iostream.h`
- [x] ostream
ostream is a low-level library and will be used for writing into file descriptors, This class provide a buffer for each output stream (4K buffer size yet) and provide `print()` function for writing into fds. `print()` and `printerr()` are designed to replace `cout` and `cerr`. Here's basic usage of them:
```c++
dio::print("Hello World\n");
std::string service_name = "boot";
unsigned load_time = 12;
dio::printerr("Dinit loaded", service_name, "in", load_time, "seconds");
```
- [ ] istream
same as ostream but for reading from file descriptors. ToDo

high-level header: `dinit-fstream.h`
- [ ] ofstream
ofstream is a high-level library and will be used for writing into files, This class open a file and create a ostream object and will provide some functions such as `open()`. ToDo
- [ ] ifstream
same as ofstream but for reading from files. ToDo

Also dinitcheck modified to use these new header(s) for catching any regression. I hope it'll be on right track. Regards

### DIFF
--- a/src/dinitcheck.cc
+++ b/src/dinitcheck.cc
@@ -1,5 +1,5 @@
 #include <algorithm>
-#include <iostream>
+#include "dinit-iostream.h"
 #include <fstream>
 #include <cstring>
 #include <string>
@@ -71,7 +71,7 @@ static bool errors_found = false;
 
 int main(int argc, char **argv)
 {
-    using namespace std;
+    using namespace dio;
 
     service_dir_opt service_dir_opts;
     bool am_system_init = (getuid() == 0);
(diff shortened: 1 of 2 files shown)


src/dinitcheck.cc
```diff
@@ -88,21 +88,21 @@ int main(int argc, char **argv)
                         service_dir_opts.set_specified_service_dir(argv[i]);
                     }
                     else {
-                        cerr << "dinitcheck: '--services-dir' (-d) requires an argument" << endl;
```
mobin-2008, line 91: We need to have "endl"?

src/includes/dinit-iostream.h
```diff
@@ -0,0 +1,130 @@
...
+constexpr unsigned BUF_SIZE = 4096; // MAXMUM size of buffers
+
+namespace dio {
+
+template <int fd> class ostream
+{
+    private:
+
+        cpbuffer<BUF_SIZE> buf; // Main buffer
+        bool stdio_sync = false; // Make sure message writed into dest
```
mobin-2008, line 19: syncing with stdio is disabled by default, Is it fine?

davmac314:
Please make your own judgement and we can discuss if necessary when the PR is complete. You did ask to take on this issue so it is your responsibility! If there are questions you really can't resolve, you can ask them in the [issue](https://github.com/davmac314/dinit/issues/240) so we have a reasonable record of the decisions there.

I would prefer not to have draft PRs - they are distracting (for me, due to all the notifications) and not a good way to discuss technical details. If you need to show some example of code in a discussion, you can link to one of your own branches for example. I'm going to unsubscribe from this PR, you will need to notify me when it is ready.

src/dinitcheck.cc
davmac314: The `endl` there looks superfluous. It might be needed elsewhere, if the buffer needs to be flushed.

src/includes/dinit-iostream.h
davmac314: is this supposed to be the equivalent of `ios_base::sync_with_stdio`? The comment doesn't match that. What do you mean by "sync with stdio"?

mobin-2008:
I will create a new PR when it mostly done

PR #248, opened 2023-09-28
igr-runner: Better log visibility with `DEBUG=1`

mobin-2008:

### DIFF
--- a/src/igr-tests/igr-runner.cc
+++ b/src/igr-tests/igr-runner.cc
@@ -32,7 +32,7 @@ int main(int argc, char **argv)
         std::string prog_path = "./run-test.sh";
         char * const p_argv[2] = { const_cast<char *>(prog_path.c_str()), nullptr };
 
-        std::cout << test_dir << "... ";
+        std::cout << std::endl << "### " << test_dir << "... ";
 
         // "Use posix_spawn", they said. "It will be easy", they said.
 


mobin-2008:
Alpine-latest_build (armv6) seems unrelated but I will re-run the tests again.

mobin-2008:
Hmm, Duckerhub seems to be experiencing some major issues

src/igr-tests/igr-runner.cc
```diff
@@ -32,7 +32,7 @@ int main(int argc, char **argv)
         std::string prog_path = "./run-test.sh";
         char * const p_argv[2] = { const_cast<char *>(prog_path.c_str()), nullptr };
 
-        std::cout << test_dir << "... ";
+        std::cout << std::endl << "### " << test_dir << "... ";
```
davmac314, line 35: This changes the output regardless of DEBUG value

mobin-2008: Right but checking for DEBUG looks unnecessary I guess

mobin-2008: Do you mean to make it to check DEBUG or change the commit message? (I'm finding this commit msg isn't suitable now, sorry)

davmac314: Ok so you meant change the output generally so that it easier to see debugging messages as each test runs?

The problem is, I don't want the output to change (especially when DEBUG is not set). It would then be inconsistent with the unit tests and it's just extra noise. The usual case is to run the tests without DEBUG set.

Even for CI, we shouldn't really need to use this extra DEBUG setting. If it's necessary to see the output then it should be logged somewhere (and the CI can upload the logfile as an artifact if the tests fail for example). For successful runs it's currently just producing a lot of noise.

mobin-2008: I close this PR to redesign the whole `DEBUG` functionality (Writing into files instead of stdout,err mainly).

Issue #179, opened 2023-04-22
unifying the data types for socket communication


james-knippes:
dinit and dinitctl define some  data types used in socket communication in different places/headers.
E.g.:
- `control.cc` gets the definition of  `handle_t` from the `control_conn_t` class in  `control.h`
- `dinitctl.cc` gets it's definition of `handle_t` directrly from from `dinit-client.h`

If the type used for socket communication has to be changed this has to be done/tracked in two different places.
I would suggest creating the a common header for types used in the communication protocol.  Maybe pull them all in `includes/control-datatypes.h` and go through the code moving the used datatypes there?
I could create a mr if there are no general objections?

davmac314:
I'm not opposed, thanks.

mobin-2008:
Addressed by #250

PR #250, opened 2023-10-23
Unify socket datatypes

james-knippes:
Hey, I've started to work on getting the datatypes used in the socket communication unified. (See #179 ) The types are now defined in control-datatypes.h .

The process led to a little bit of refactoring of the control commands in control-cmds.h .

Before I go on with the other datatypes and control commands I would appriciate some feedback if you have the time @mobin-2008 or @davmac314 .

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -6,6 +6,7 @@
 #include "control.h"
 #include "service.h"
 #include "proc-service.h"
+#include "control-datatypes.h"
 
 // Server-side control protocol implementation. This implements the functionality that allows
 // clients (such as dinitctl) to query service state and issue commands to control services.
@@ -14,6 +15,9 @@
 // 1 - dinit 0.16 and prior
 // 2 - dinit 0.17 (adds DINIT_CP_SETTRIGGER, DINIT_CP_CATLOG, DINIT_CP_SIGNAL)
 // 3 - (unreleased) (adds DINIT_CP_QUERYSERVICEDSCDIR)
+
+// common communication datatypes
+using namespace dinit_ctypes;
 
 namespace {
     // Control protocol minimum compatible version and current version:
@@ -1152,7 +1156,7 @@ bool control_conn_t::query_load_mech()
     }
 }
 
-control_conn_t::handle_t control_conn_t::allocate_service_handle(service_record *record)
+handle_t control_conn_t::allocate_service_handle(service_record *record)
 {
     // Try to find a unique handle (integer) in a single pass. Since the map is ordered, we can search until
     // we find a gap in the handle values.
--- a/src/control.cc
+++ b/src/control.cc
@@ -928,7 +928,7 @@ bool control_conn_t::process_set_trigger()
     }
 
     handle_t handle;
-    char trigger_val;
+    trigger_val_t trigger_val;
 
     rbuf.extract(&handle, 1, sizeof(handle));
     rbuf.extract(&trigger_val, 1 + sizeof(handle), sizeof(trigger_val));
--- a/src/control.cc
+++ b/src/control.cc
@@ -45,8 +45,8 @@ bool control_conn_t::process_packet()
     cp_cmd pktType = (cp_cmd)rbuf[0];
     if (pktType == cp_cmd::QUERYVERSION) {
         // Responds with:
-        // DINIT_RP_CVERSION, (2 byte) minimum compatible version, (2 byte) actual version
-        char replyBuf[] = { DINIT_RP_CPVERSION, 0, 0, 0, 0 };
+        // cp_rply::CVERSION, (2 byte) minimum compatible version, (2 byte) actual version
+        char replyBuf[] = { (char)cp_rply::CPVERSION, 0, 0, 0, 0 };
         memcpy(replyBuf + 1, &min_compat_version, 2);
         memcpy(replyBuf + 3, &cp_version, 2);
         if (!queue_packet(replyBuf, sizeof(replyBuf))) return false;
@@ -81,7 +81,7 @@ bool control_conn_t::process_packet()
             auto sd_type = static_cast<shutdown_type_t>(rbuf[1]);
 
             services->stop_all_services(sd_type);
-            char ackBuf[] = { DINIT_RP_ACK };
+            char ackBuf[] = { (char)cp_rply::ACK };
             if (! queue_packet(ackBuf, 1)) return false;
 
             // Clear the packet from the buffer
@@ -130,7 +130,7 @@ bool control_conn_t::process_packet()
     }
 
     // Unrecognized: give error response
-    char outbuf[] = { DINIT_RP_BADREQ };
+    char outbuf[] = { (char)cp_rply::BADREQ };
     if (!queue_packet(outbuf, 1)) return false;
     bad_conn_close = true;
     return true;
@@ -151,7 +151,7 @@ bool control_conn_t::process_find_load(cp_cmd pktType)
     rbuf.extract((char *)&svcSize, 1, 2);
     if (svcSize <= 0 || svcSize > (1024 - 3)) {
         // Queue error response / mark connection bad
-        char badreqRep[] = { DINIT_RP_BADREQ };
+        char badreqRep[] = { (char)cp_rply::BADREQ };
         if (! queue_packet(badreqRep, 1)) return false;
         bad_conn_close = true;
         return true;
@@ -171,7 +171,7 @@ bool control_conn_t::process_find_load(cp_cmd pktType)
     rbuf.consume(chklen);
     chklen = 0;
     
-    char fail_code = DINIT_RP_NOSERVICE;
+    cp_rply fail_code = cp_rply::NOSERVICE;
 
     if (pktType == cp_cmd::LOADSERVICE) {
         // LOADSERVICE
@@ -180,17 +180,17 @@ bool control_conn_t::process_find_load(cp_cmd pktType)
         }
         catch (service_description_exc &sdexc) {
             log_service_load_failure(sdexc);
-            fail_code = DINIT_RP_SERVICE_DESC_ERR;
+            fail_code = cp_rply::SERVICE_DESC_ERR;
         }
         catch (service_not_found &snf) {
             log(loglevel_t::ERROR, "Could not load service ", snf.service_name, ": ",
                     snf.exc_description);
-            // fail_code = DINIT_RP_NOSERVICE;   (already set)
+            // fail_code = cp_rply::NOSERVICE;   (already set)
         }
         catch (service_load_exc &slexc) {
             log(loglevel_t::ERROR, "Could not load service ", slexc.service_name, ": ",
                     slexc.exc_description);
-            fail_code = DINIT_RP_SERVICE_LOAD_ERR;
+            fail_code = cp_rply::SERVICE_LOAD_ERR;
         }
     }
     else {
@@ -199,7 +199,7 @@ bool control_conn_t::process_find_load(cp_cmd pktType)
     }
     
     if (record == nullptr) {
-        std::vector<char> rp_buf = { fail_code };
+        std::vector<char> rp_buf = { (char)fail_code };
         if (! queue_packet(std::move(rp_buf))) return false;
         return true;
     }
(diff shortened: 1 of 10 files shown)


davmac314:
Looks good to me so far, I don't have any concerns other than minor things I'll comment.

src/control.cc
```diff
@@ -373,13 +378,17 @@ bool control_conn_t::process_start_stop(int pktType)
...
+        case control_cmd::CP_RELEASESERVICE:
             // remove required mark, stop if not required by dependents
             if (do_pin) service->pin_stop();
             service->stop(false);
             services->process_queues();
             if (service->get_state() == service_state_t::STOPPED) ack_buf[0] = DINIT_RP_ALREADYSS;
             break;
+        default:
+            // can't get here (hopefully)
+            // supreses compiler warning about unused  (-Wswitch)
```
davmac314, line 390: Please keep the comment to a minimum - since this is just for suppressing a warning, as little noise as possible is preferable. I suggest:
```
// avoid warning for unhandled values in switch/case
```

src/includes/control-datatypes.h
```diff
@@ -0,0 +1,14 @@
+#ifndef DINIT_CONTROL_DATATYPES_H
+#define DINIT_CONTROL_DATATYPES_H
+
+#include <cstdint>
+
+namespace dinit_ctypes {
```
davmac314, line 6: `ctypes` sounds a bit too close to the existing character types, maybe `cptypes` or `cp_types` (for "command protocol types")?

davmac314:
Just the two comments, otherwise looks good, thanks!

src/control.cc
james-knippes: done.

src/includes/control-datatypes.h
james-knippes: seems reasonable to avoid confusion. it's dinit_cptypes now

james-knippes:
Thanks for the Feedback @davmac314 . I've continued refactoring in the other datatypes used in communication in the same way. I think by now i covered most of them.

To be honest, I have some difficulties reading/understanding the protocol related code. It feels a  bit 'fragile' with all the hardcoded values/lengths in different places. Most of it should be avoidable with some rework using sizeof/types. Or we could move the static parts of the protocol into standard layout type structs?

davmac314:
> To be honest, I have some difficulties reading/understanding the protocol related code. It feels a bit 'fragile' with all the hardcoded values/lengths in different places. Most of it should be avoidable with some rework using sizeof/types. Or we could move the static parts of the protocol into standard layout type structs?

It is slightly tricky code. A lot of it already uses `sizeof` where it can (except for when it would be `sizeof(char)`, it will just use 1). I certainly am happy if you change code which is using fixed sizes > 1 to use a `sizeof` instead.

Using regular structs is a bit less simple, since you'd need to make sure they were exactly compatible with the current protocol. Standard C++ has no way to specify a packed struct as far as I know, so this would be only be possible where the packet format already matches standard layout. If you could keep within those constraints I'm happy for such changes, but I think that should be part of a second change set (not part of this PR).

src/includes/control-datatypes.h
```diff
@@ -0,0 +1,20 @@
...
+    // A mapping between service records and their associated numerical identifier
+    // used in communction
+    using handle_t = uint32_t;
+    using trigger_val_t = uint8_t;
+    using cp_cmd_t = uint8_t;
+    using cp_rply_t = uint8_t;
+    using cp_info_t = uint8_t;
+    using srvname_len_t = uint16_t;
+    using envvar_len_t = uint16_t;
+    using sig_num_t = int32_t;
```
davmac314, line 16: This one (`sig_num_t`) isn't correct, a signal number is an `int` and not guaranteed to be 32-bit.

src/control.cc
```diff
@@ -37,34 +42,34 @@ bool control_conn_t::process_packet()
...
 
-    int pktType = rbuf[0];
-    if (pktType == DINIT_CP_QUERYVERSION) {
+    cp_cmd pktType = (cp_cmd)rbuf[0];
+    if (pktType == cp_cmd::QUERYVERSION) {
         // Responds with:
-        // DINIT_RP_CVERSION, (2 byte) minimum compatible version, (2 byte) actual version
-        char replyBuf[] = { DINIT_RP_CPVERSION, 0, 0, 0, 0 };
+        // cp_rply::CVERSION, (2 byte) minimum compatible version, (2 byte) actual version
+        char replyBuf[] = { (cp_rply_t)cp_rply::CPVERSION, 0, 0, 0, 0 };
```
davmac314, line 49: Casts should be to the destination type, which would be `char` here, if needed at all.

`(cp_rply_t)cp_rply::CPVERSION` is superfluous - the type of `cp_rply::CPVERSION` is already `cp_reply_t`.

src/control.cc
```diff
@@ -76,7 +81,7 @@ bool control_conn_t::process_packet()
             auto sd_type = static_cast<shutdown_type_t>(rbuf[1]);
 
             services->stop_all_services(sd_type);
-            char ackBuf[] = { DINIT_RP_ACK };
+            char ackBuf[] = { (cp_rply_t)cp_rply::ACK };
```
davmac314, line 84: Similarly unnecessary/incorrect cast here, and a bunch throughout, I think.

src/control.cc
```diff
@@ -160,41 +165,41 @@ bool control_conn_t::process_find_load(int pktType)
...
     }
     else {
         // FINDSERVICE
-        record = services->find_service(serviceName.c_str());
+        record = services->find_service(service_name.c_str());
     }
     
     if (record == nullptr) {
-        std::vector<char> rp_buf = { fail_code };
+        std::vector<char> rp_buf = { (cp_rply_t)fail_code };
```
davmac314, line 202: (This one is causing build failure on Mac, clang is less lenient than gcc)

src/includes/control-cmds.h
```diff
@@ -1,148 +1,152 @@
 #ifndef DINIT_CONTROL_CMDS_H_INCLUDED
 #define DINIT_CONTROL_CMDS_H_INCLUDED 1
 
+#include "control-datatypes.h"
+
 // Dinit control command packet types
 
 // Requests:
+enum class cp_cmd :dinit_cptypes::cp_cmd_t {
```
davmac314, line 9: formatting nit: can you put spaces around `:` to match style elsewhere? i.e
```
enum class cp_cmd : dinit_cptypes::cp_cmd_t {
```

davmac314:
Had a quick look through and added some comments, mostly looking good though, thanks!

src/includes/control-cmds.h
james-knippes: done.

src/control.cc
james-knippes: The underlying Problem seems to be, that when variables in the initilizer list are non static and the types are not the same then there is narrowing involved (which I don't fully understand). So casting to (char) works, but casting to (cp_rply_t)/(unsigned char) does not, as it's not clear if the value fits into char. So I've opted for casting to (char) in this case.

src/control.cc
james-knippes: I think a cast is actually necessary here, because cp_rply::CPVERSION is of type cp_rply and not cp_rply_t, as it's an enum class.

I've looked it up and narrowing is also applied on static variables in initializer lists. The difference between casting to char or cp_rply_t type in this case would mostly be, that the compiler complains as soon as the value would be out of range for the target type (char, so >127 in this case). So I agree casting to (char) is probably best here.

src/includes/control-datatypes.h
james-knippes: Yes, I'm aware that it is not guaranteed to be 32-bit.
The question here would be do we want the protocol to be stable (even across architectures/systems) so it could be used over system boundaries (e.g. over network and/or forwarded to container/vm) ? In that case I would propose to leave it be fixed to 32-bit for the protocol and cast to int in the application. When there is a system, where int is not 32-bit then we may need to handle the conversion in dinitctl differently with an additional check, warning or error.

james-knippes:
> It is slightly tricky code. A lot of it already uses `sizeof` where it can (except for when it would be `sizeof(char)`, it will just use 1). I certainly am happy if you change code which is using fixed sizes > 1 to use a `sizeof` instead.
Yes i will create an new issue/PR for that to no clutter this PR to much.

> Using regular structs is a bit less simple, since you'd need to make sure they were exactly compatible with the current protocol. Standard C++ has no way to specify a packed struct as far as I know, so this would be only be possible where the packet format already matches standard layout. If you could keep within those constraints I'm happy for such changes, but I think that should be part of a second change set (not part of this PR).
Ah I didn't thought about packed structs not being standard c++. As we are currently only targeting compilers, which both (gcc and clang) support __attribute__((packed)) for structs we may use it? Could discuss this in a new Issue if you don't rule it out. Maybe also discuss other possible ways to optimize the protocol (handling).

src/control.cc
davmac314: Right, you're correct, I misread it - still, casting to char is probably best, as you say.

james-knippes: I've changed all relevant casts to (char).

james-knippes:
The remaining issues should be fixed now. Let me know if i missed something. I will squash the fixes into the related commits later.

src/includes/control-datatypes.h
davmac314: I've never had the goal of having the protocol be usable over system boundaries with heterogeneous systems. That requires more than just consistently sized values; you need to ensure consistent byte order, as well, and no struct declaration will do that for you.

I know that for the most part `int32_t` will work just fine - 99% of the time it will be the case that `sizeof(int) == sizeof(int32_t)` - and the remainder of the time signal numbers will probably not be bigger than 2^31 anyway. However, although it's a bit arbitrary, "works on any POSIX system" _is_ a design goal and POSIX just says that signal numbers are `int`.

james-knippes: So we should be in the clear here, as dinitctl handles the out_of_range exception if the specified signal number does not fit into a sig_num_t/int32_t. I've also added a cast from int32_t to int in control.cc, to make it more explicit. So when there is a system with different sizes it should not break, we just don't support signal numberrs >32bit atm.

davmac314: I may have been unclear. Sorry, but no - the signal number type needs to be `int`. That's what POSIX declares it as.

> as dinitctl handles the out_of_range exception if the specified signal number does not fit into a sig_num_t/int32_t

dinitctl uses `std::stoi` to convert the signal number and that returns `int`. It only throws `out_of_range` if the number doesn't fit into an `int` - it doesn't know anything about `sig_num_t`/`int32_t`. So if `int` is larger than 32-bits there's no checking done in dinitctl that the provided signal number will fit in `int32_t`. Just assigning the result to `int32_t` will silently truncate it. (Or am I missing something?)

> we just don't support signal numberrs >32bit atm

That would be a regression introduced with this change; currently, we do in fact support signal numbers >32 bit.

Everything else in the PR looks good!

src/includes/control.h
```diff
@@ -15,6 +15,7 @@
 #include <control-cmds.h>
 #include <service-listener.h>
 #include <cpbuffer.h>
+#include "control-datatypes.h"
```
mobin-2008, line 18: Hi, I reviewd your work and it looks good, unless that `control-datatypes.h` should included like this, I guess (with `< >` not `" "`):
```c++
#include <control-datatypes.h>
```

src/includes/control-datatypes.h
james-knippes: Ah.. now I get it. I've reverted sig_num_t to be int. And I've checked again, that at least in the signal protocol code sizeof is used consequently, so it doesn't break the protocol on systems where sizeof(int) != 4.

src/includes/control.h
james-knippes: Hey, nice catch thanks. Functionally it should not make a difference in this case. But I've changed the includes of control-datatypes.h so they are consistent with the others in the corresponding file.

mobin-2008 (approved):
Looks good, Thanks for contribution 😁

davmac314:
Thanks! Merged!

PR #257, opened 2023-11-06
dinit monitor execute command for initial status

james-knippes:
I've refactored dinit-monitor a little bit to execute commands for the initial service state. Also added options to specify strings to be substituted per service state.

### DIFF
--- a/doc/manpages/dinit-monitor.8.m4
+++ b/doc/manpages/dinit-monitor.8.m4
@@ -38,12 +38,20 @@ This option determines the default path to the control socket used to communicat
 (it does not override the \fB\-p\fR option).
 .TP
 \fB\-i\fR, \fB\-\-initial\fR
-Issue the specified command additionally for the initial status of a service.
-Otherwise, the command is only executed when the status changes for the first time.
+Issue the specified command additionally for the initial status of the services (when \fBdinit\-monitor\fR is started).
+Without this option, the command is only executed whenever service status changes.
 .TP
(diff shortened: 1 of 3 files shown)


davmac314:
Hey @james-knippes this is still set to draft, are you wanting review at this stage? From a quick look the only things I can point out are:

- Needs the man page for dinit-monitor to be updated
- I think it would be better from a user experience point-of-view to load all services specified first, then do the initial notification on each one. That way if one fails to load (wrong service name etc) you get an immediate error without getting notifications for the others.

james-knippes:
Hey @davmac314,
yes thanks for the review. I've updated the man page and separated service loading and the initial status commands into two steps. Looks good now I think.

davmac314:
I will look at this properly when I get a chance - may not be until the weekend.

davmac314:
Hi @james-knippes , I pushed a commit with formatting fixes and some rewording. Please check it and see if you are happy with it. If so I'm happy to merge.

Unfortunately the commit is not showing in the github UI here in the PR, but it is visible on [your branch](https://github.com/james-knippes/dinit/tree/feature_dinit-monitor-initial-status). I can merge it manually if necessary.

james-knippes:
Thanks @davmac314 , sounds way better now. Appreciate you also including explanations for the formatting changes/decisions in the commit. I happy with it.

Also for me the commit is not missing from the github UI. At least in this pr it is shown in the conversation/history for me.

davmac314:
Seems there was just a delay in the UI showing the commit, it appears for me now too. Merged! Thanks.

PR #260, opened 2023-12-10
dinitcheck: account for service working directory for commands

q66:


src/dinitcheck.cc
```diff
@@ -429,9 +432,19 @@ service_record *load_service(service_set_t &services, const std::string &name,
 
     settings.finalise(report_err, renvmap, report_err, resolve_var);
 
+    if (!settings.working_dir.empty()) {
+        service_wdir = settings.working_dir;
+    }
+    dirfd = open(service_wdir.c_str(), O_DIRECTORY | O_PATH);
```
mobin-2008, line 438: Hi.

The problem with this PR is that `O_PATH` is not defined by POSIX and it means there is no guarantee about its availability (e.g MacOS ci is failing due to this).

I guess opening with `O_RDONLY` is better option but It makes into problem when directory read access isn't permitted but execute access is permitted.

Ref: https://pubs.opengroup.org/onlinepubs/9699919799/

mobin-2008 (changes requested):

src/dinitcheck.cc
mobin-2008: I'm not sure opening directory with just `O_DIRECTORY` is safe. Is it?

q66: no idea what you mean?

mobin-2008: Nevermind, I didn't see O_RDONLY

davmac314 (approved):
LGTM

PR #264, opened 2023-12-18
Add $XDG_CONFIG_HOME/dinit.d/ to default user service directories

### LINKED ISSUE #222: `XDG_CONFIG_HOME` support for user services.
i can pass `--services-dir`, but it doesnt seem like a good option if `dinit` is launched by `turnstyled`, and its just a nice default to have.

into-the-v0id:
Dinit's default user service directory is `$HOME/.config/dinit.d/`. It does not respect / comply with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). This PR adds `$XDG_CONFIG_HOME/dinit.d/` as an additional default user service directory (alongside `$HOME/.config/dinit.d/` for backwards compatibility).

When searching for service definition files as a user instance, the following cases are possible:
- no `$XDG_CONFIG_HOME` is set
  -> only search `$HOME/.config/dinit.d/`
- `$XDG_CONFIG_HOME` is set to `$HOME/.config`
  -> `$HOME/.config/dinit.d/` will be searched twice
- `$XDG_CONFIG_HOME` is set to a custom directory
  -> search `$XDG_CONFIG_HOME/dinit.d/` and then `$HOME/.config/dinit.d/`

As far as I can tell, this should provide full backwards compatibility: The `$HOME/.config/dinit.d/` directory is still always searched and it's unreasonable to assume that the user already has a custom `$XDG_CONFIG_HOME` with a `dinit.d/` directory.

Feel free to ask for any changes to this PR or to reject it :)

Fixes #222

### DIFF
--- a/src/options-processing.cc
+++ b/src/options-processing.cc
@@ -56,8 +56,20 @@ void service_dir_opt::build_paths(bool am_system_init)
                 std::memcpy(service_dir_w + user_home_len, "/.config/dinit.d", dinit_d_len);
                 service_dir_w[full_len - 1] = 0;
 
-                service_dir_paths.emplace_back(service_dir_w, /*dyn_allocd=*/true);
-                home_service_dir_set = true;
+                bool user_home_path_already_added = false;
+                for (auto &path : service_dir_paths) {
+                    if (strcmp(path.get_dir(), service_dir_w) == 0) {
+                        user_home_path_already_added = true;
+                        break;
+                    }
+                }
+
+                if (!user_home_path_already_added) {
+                    service_dir_paths.emplace_back(service_dir_w, /*dyn_allocd=*/true);
+                    home_service_dir_set = true;
+                } else {
+                    delete[] service_dir_w;
+                }
             }
         }
 
--- a/src/options-processing.cc
+++ b/src/options-processing.cc
@@ -32,6 +32,20 @@ void service_dir_opt::build_paths(bool am_system_init)
 
         /* service directory name */
         if (!am_system_init) {
+            const char * xdg_config_home = getenv("XDG_CONFIG_HOME");
+            if (xdg_config_home != nullptr && strlen(xdg_config_home) != 0) {
+                size_t xdg_config_home_len = strlen(xdg_config_home);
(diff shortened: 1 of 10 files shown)


davmac314:
Thanks for submitting this. Could you add a check whether `$XDG_CONFIG_HOME` is actually `$HOME/.config` and not add the directory a second time if so? I know that adding it twice basically has no effect, but it still seems like the wrong thing to do.

into-the-v0id:
Sure thing - I've added an additional check to only add `$HOME/.config/dinit.d/` to the default user service directories if it is not set already (through `$XDG_CONFIG_HOME/dinit.d/`)

src/options-processing.cc
```diff
@@ -42,8 +56,20 @@ void service_dir_opt::build_paths(bool am_system_init)
                 std::memcpy(service_dir_w + user_home_len, "/.config/dinit.d", dinit_d_len);
                 service_dir_w[full_len - 1] = 0;
 
-                service_dir_paths.emplace_back(service_dir_w, /*dyn_allocd=*/true);
-                home_service_dir_set = true;
+                bool user_home_path_already_added = false;
+                for (auto &path : service_dir_paths) {
```
davmac314, line 60: Note that due to the `service_dirs.empty()` check on line 30 this `for` loop is really overkill - there's only one other possible directory to check against, i.e. the one that was possibly just added. But that's ok. I might tweak it a bit after merging.

davmac314:
Merged, thanks!

src/options-processing.cc
into-the-v0id: Ok. I thought, that `service_dirs` are the dirs manually specified via the `--services-dir` flag. So assuming we dont have that flag and call `service_dir_opt::build_paths()` multiple times (for whatever reason), we would end up with more than one item in `service_dir_paths`. However in this case, all the other directories (`/etc/dinit.d`, `$XDG_CONFIG_HOME/dinit.d`, etc.) would be added twice anyway and this case probably does not even occur. So only checking for the first item (if it exists) should probably be fine.

Sorry - I'm kind of new to C++ stuff and am just trying to understand it :)

Issue #266, opened 2023-12-24
macOS / Darwin build needs different `SBINDIR` and `MANDIR` 


aidan-gibson:
`/sbin` and `/usr/share/man` are protected by [SIP](https://support.apple.com/en-us/102149#:~:text=System%20Integrity%20Protection%20is%20a,and%20folders%20on%20your%20Mac.). I built very easily by changing `mconfig`, so not a big deal, but just letting ya know a macOS / Darwin user with SIP enabled will get an error if they try to build using the instructions as they currently are. (Thank you for building this btw! Really happy to be using this on macOS.)

aidan-gibson:
Ended up adding it to homebrew https://github.com/aidan-gibson/homebrew-dinit

mobin-2008:
Merged, Thanks :). Also you can add yourself to [`CONTRIBUTORS`](https://github.com/davmac314/dinit/blob/master/CONTRIBUTORS) list if you wish.

mobin-2008:
Fixed in mconfig, configure needs patch, I'm going to fix it.

davmac314:
Instructions have been updated, I'll close this. Thanks for reporting.

PR #267, opened 2023-12-24
Update mconfig.Darwin

aidan-gibson:
Solves #266 
/sbin and /usr/share/man are protected by SIP

### DIFF
--- a/configs/mconfig.Darwin
+++ b/configs/mconfig.Darwin
@@ -1,7 +1,7 @@
 # Installation path options.
 
-SBINDIR=/sbin
-MANDIR=/usr/share/man
+SBINDIR=/usr/local/sbin
+MANDIR=/usr/local/share/man
 SYSCONTROLSOCKET=/var/run/dinitctl
 
 


configs/mconfig.Darwin
```diff
@@ -1,7 +1,7 @@
 # Installation path options.
 
-SBINDIR=/sbin
-MANDIR=/usr/share/man
+SBINDIR=/usr/local/sbin
+MANDIR=/usr/local/share/man
 SYSCONTROLSOCKET=/var/run/dinitctl
```
mobin-2008, line 5: Hi. Thanks for the contribution :)

Hmm, As I read, /var is also protected by SIP. It should be changed to some where else I guess (or is /var/run a exception?)

mobin-2008 (approved):
Looks good except I have question about /var/run/

configs/mconfig.Darwin
aidan-gibson: You're correct, `/var` is protected but `/var/run` is an exception.

davmac314:
Sorry for being unable to review this myself earlier. I think this change should not have gone through, for the following reasons:

- the default build configuration is provided with the assumption of installing Dinit as a "system" init. For most systems other than (some) Linux systems this is not really feasible but it's nevertheless the default. It can be overridden for all systems via setting appropriate variables (`SBINDIR=...`) on the command line when doing the build (and `make install` step).
- SIP, which prevents the default from working at all on MacOS, can be disabled in order to install software in the "protected" directories, and it's reasonable to think that if someone wanted a "system" install of Dinit that they would do this.
- The change affects the "Darwin" configuration, which affects not just MacOS but also other systems based on Darwin, such as PureDarwin (which does not have SIP).

I think all that was really needed was to adjust the build instructions to make it more clear that `SBINDIR` will often need to be specified (perhaps even with specific mention of SIP). With that in mind, I'm going to revert this change and apply an appropriate documentation change in its place.

Apologies @aidan-gibson for the hassle. I'm assuming this won't affect homebrew installs at all. If you think what I've proposed really is the wrong solution then feel free to post here. Unfortunately your PR was merged before the usual review, which shouldn't have happened.

@mobin-2008 Please don't merge PRs before we have discussed them, in future. I value your input and your reviews but this is my project and I maintain full responsibility for it, which also means I must be the final arbiter and in this case as I have outlined, I do not agree that this should have been merged.

PR #268, opened 2023-12-24
configure: Use paths from "/usr/local" on Darwin by default

### LINKED ISSUE #266: macOS / Darwin build needs different `SBINDIR` and `MANDIR`

mobin-2008:
SIP (System Integrity Protection) is a feature introduced in OS X El Capitan which protects some paths (include "/usr`) from modifying or overwriting. This patch changes "SBINDIR" and "MANDIR" to paths from "/usr/local/" by default on MacOS.

Fixes #266 

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -208,8 +208,9 @@ DESTDIR:
 
     make DESTDIR=/some/path install
 
-The dinit executable will be put in /sbin (or rather, in $DESTDIR/sbin), which may not be on the
-path for normal users. Consider making a symbolic link to /usr/sbin/dinit.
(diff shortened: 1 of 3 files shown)


davmac314:
As noted in #267 I do not agree with this change and did not approve the equivalent change to the standard build, so it has been reverted.

PR #263, opened 2023-12-16
dinit-iostream & dinit-fstream: Custom new special-purpose libraries

mobin-2008:
These are replacement for C++ standard library iostream/fstream.

See dinit-iostream.h and dinit-fstream.h for more information and the usage.

Closes #240

`Signed-off-by: Mobin "Hojjat" Aydinfar <mobin@mobintestserver.ir>`

PR #272, opened 2024-01-07
meson: Use shell globbing for listing igr-tests

mobin-2008:
There was at least 2 new tests which doesn't run in meson builds, This PR use shell globbing for listing igr-tests which is automated.

### DIFF
--- a/src/igr-tests/meson.build
+++ b/src/igr-tests/meson.build
@@ -5,172 +5,18 @@ igr_tests_env += [
      'DEBUG=yes'
 ]
 
-# Integration tests
-basic_test_script = files(meson.current_source_dir() + '/basic/run-test.sh')
-environ_test_script = files(meson.current_source_dir() + '/environ/run-test.sh')
-environ2_test_script = files(meson.current_source_dir() + '/environ2/run-test.sh')
-psenviron_test_script = files(meson.current_source_dir() + '/ps-environ/run-test.sh')
-chainto_test_script = files(meson.current_source_dir() + '/chain-to/run-test.sh')
-forcestop_test_script = files(meson.current_source_dir() + '/force-stop/run-test.sh')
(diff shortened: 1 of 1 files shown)


davmac314 (changes requested):
This modification involves invoking two separate shell commands and is clearly fighting against meson's limitations as well as using a brittle hack to identify tests. As we're approaching 1.0 release I think it's not the time to introduce any hacks in the build.

As I've suggested before, the right way to run the integration tests is via `igr-runner` itself. If you really want to run the tests individually from the meson build then the best solution is probably to make changes to `igr-runner` necessary to allow it to list available tests and to run an individual test, and then have the meson build use that functionality.

The change to use `igr-runner` needs to happen at some point. The integration tests will be re-engineered soon. I warned about this a while back, most recently in August, see https://github.com/davmac314/dinit/pull/212.

> I would still prefer to run tests via igr-runner. As we get closer to 1.0 I plan to re-work the integration tests significantly and updating the meson build is going to be difficult to sustain.

If you don't want to do that now, then stick to listing the individual tests, not a hack which lists directories via a shell command greps in the source of igr-runner to try to identify tests. But it will most likely be necessary to switch to igr-runner at some point regardless.

mobin-2008:
So I'm going to change this PR to add these new tests. I will open a issue to discuss about re-engineering `igr-runner` (and probably working on it after #263 got merged).

davmac314:
I would like to do a release this weekend. Will you be able to update this PR by then? I can do it myself if you prefer

mobin-2008:
> I would like to do a release this weekend. Will you be able to update this PR by then (by Monday)? I can do it myself if you prefer

I'm not sure to have free time, thanks :)

davmac314:
I have added the missing integration tests to the meson build in a separate commit

PR #283, opened 2024-01-26
configure: use `try_optional_cxx_argument()` when flag is not required for less confusion

### LINKED ISSUE #281: configure: `-fsanitize=address,undefined` flag test is incorrectly not marked as optional!
https://github.com/davmac314/dinit/blob/17086100eae4912e87101b30e4844e7c52861693/configure#L78-L81
https://github.com/davmac314/dinit/blob/17086100eae4912e87101b30e4844e7c52861693/configure#L412-L413

As you see `try_cxx_argument()` function can return 1 when provided flag is not supported; for all optional flags result of that function should be ignored by ` || :`:
https://github.com/davmac314/dinit/blob/17086100eae4912e87101b30e4844e7c52861693/configure#L353-L357

Regerestion from: 77e60bfff13f31f190d865aee492360375827f84 (Changes behavior of that function) + d94debaa4ba7c835b4902291644231c2e878197c (Doesn't ignore `try_cxx_argument()` function result)

Regards

mobin-2008:
Previously, All flags get checked through try_cxx_argument() which can return 1 on unsupported flag and its return should be ignored but it's made into confusion.

See https://github.com/davmac314/dinit/issues/281 for a example.

This commit creates a new function that will ignore try_cxx_argument() return by default and this commit replaces all uses of try_cxx_argument() on optional flags with this new function.

Fixes #281 

### DIFF
--- a/configure
+++ b/configure
@@ -62,10 +62,12 @@ cxx_works()
 }
 
 # Test if the compiler accepts an argument (for compilation and link); if so add it to named variable.
+# NOTE: This function will return 1 on unsupported flag, use try_optional_cxx_argument() when
+# flag is not required or check its return.
 # $1 - the name of the variable to potentially add the argument to
 # $2 - the argument to add
 # CXX - the name of the compiler
-# CXXFLAGS, CXXFLAGS_EXTRA, LDFLAGS, LDFLAGS_EXTRA - any predetermined flags 
+# CXXFLAGS, CXXFLAGS_EXTRA, LDFLAGS, LDFLAGS_EXTRA - any predetermined flags
 try_cxx_argument()
 {
     info Checking whether the compiler accepts "$2"...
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -81,6 +83,12 @@ try_cxx_argument()
     fi
 }
 
+# Same as try_cxx_argument but doesn't return 1 on unsupported flag
+try_optional_cxx_argument()
+{
+    try_cxx_argument $1 $2 || :
```
davmac314, line 89: These (`$1` and `$2`) should really be quoted (`"$1" "$2"`) or they are prone to word-splitting. It's probably not a problem with current uses, but it's best practice.

davmac314 (approved):
It's ok as is for now.

davmac314 (approved):

PR #299, opened 2024-02-22
Update CONTRIBUTORS

iacore:
:thinking: Maybe I should write `Code, documentation` but I only changed one word.


davmac314 (approved):

davmac314:
> 🤔 Maybe I should write `Code, documentation` but I only changed one word.

Ironically, that's not very _consistent_ with the existing list! :laughing:

Issue #305, opened 2024-03-16
CI builds failing with odd error


davmac314:
Tests are starting to fail, even when run on commits that they have passed on successfully:

```
./proctests
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer:DEADLYSIGNAL
```

See:
Successful: https://github.com/davmac314/dinit/actions/runs/8221498110/job/22481929393
Failing: https://github.com/davmac314/dinit/actions/runs/8304831131/job/22730962409

These are both builds of revision 15e4b0c

davmac314:
@mobin-2008  Not sure if there's anything you can do, but we can keep the issue open to track it.

Worst case, we can disable builds on that architecture/combo for a while.

davmac314:
I've done a manual build on the same architecture (Debian Bullseye x86-64) and it was successful. The problem goes back a long way so I do suspect it's a system issue, but it's really strange. Seems like AddressSanitizer itself goes into an infinite loop or recursion.

PR #306, opened 2024-03-16
Draft: CI: Update Debian-bullseye to Debian-bookworm

mobin-2008:
Let's see how it's going.

I don't tag this as fix for #305 but If it's OK, we probably hit an debian bullseye specific bug or something.

### DIFF
--- a/.github/workflows/meson_ci.yml
+++ b/.github/workflows/meson_ci.yml
@@ -38,12 +38,12 @@ on:
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


mobin-2008:
Hmm, It's really weird, Meson build are OK and Make ones are not. Kinda interesting.

mobin-2008:
I think debian-bullseye is enough for our CI, Let's close this up.

Issue #303, opened 2024-03-11
System-wide location for user services


realroot2185:
From man dinit-service:
> SYNOPSIS
>        /etc/dinit.d/service-name, $XDG_CONFIG_HOME/dinit.d/service-name
> 

A system-wide location for user services could be very useful.

For example when installing some DE with user services a package could install them in ` /etc/dinit.d/user/` folder so that it will work out of the box.
Or if you install just some service e.g. the Artix package `galaxy/pipewire-dinit 20230930-2`.

Then users can make their version in `$XDG_CONFIG_HOME/dinit.d/` to override system one.

To avoid problems all user services file would go there.
E.g. scripts folder would be: `/etc/dinit.d/user/scripts`.

mobin-2008:
For reference Chimera Linux already does that:
https://github.com/chimera-linux/turnstile/blob/00fd0b1ad7b5fd262bb83c75cb463ad32b1940c9/backend/dinit#L78-L88

davmac314:
As Mobin points out this is already possible, it just needs the dinit user instance to be started with the right command line to add the services directory (`/etc/dinit.d/user`) as well the user's own (normally `~/.config/dinit.d`). That should be managed by whatever you have in place to launch the dinit user instances (eg Turnstile does this out of the box).

Anything I'm missing or can we close this?

realroot2185:
I think that as dinit looks in `/etc/dinit.d/service-name` and `$XDG_CONFIG_HOME/dinit.d/service-name` it should look in `/etc/dinit.d/user` out of the box.
And not relying on something else.

For example in sxmo I use dinit for user services and I just use `dinit -u`.

davmac314:
As per the readme, we are not accepting feature requests except in certain circumstances, unless you are willing to work on it yourself. Is that the case?

realroot2185:
In my opinion this is something that should be in dinit.

It thinks that it is not hard to add especially for dinit devs.
Is that maybe a good circumstance?

I can try to look but I do not know.

Here?
https://github.com/davmac314/dinit/blob/8f5268746a7138bf31de9e4d2d059cd02a33fe11/src/options-processing.cc#L28

davmac314:
> In my opinion this is something that should be in dinit.
> It thinks that it is not hard to add especially for dinit devs.
Is that maybe a good circumstance?

[...]

PR #307, opened 2024-03-21
System-wide location for user services

realroot2185:
The service in `$HOME/.config/dinit.d` overrides `/etc/dinit.d/user` version.
And services in the latter work.

### DIFF
--- a/src/options-processing.cc
+++ b/src/options-processing.cc
@@ -75,6 +75,7 @@ void service_dir_opt::build_paths(bool am_system_init)
         }
 
         done_user_home:
+        service_dir_paths.emplace_back("/etc/dinit.d/user", /*dyn_allocd=*/false);
 
         if (!home_service_dir_set) {
             service_dir_paths.emplace_back("/etc/dinit.d", /*dyn_allocd=*/false);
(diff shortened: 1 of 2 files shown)


mobin-2008:
Closes #303

davmac314 (commented):
Thanks for submitting this. A few points:

- On reflection, shouldn't the following directories also be added, for consistency with system service directories: `/usr/local/lib/dinit.d/user`, `/usr/lib/dinit.d/user`? (As it happens these are also currently supported for user services in Chimera).
- Normally commit messages should be descriptive, not just "update file xxx". This is not hugely important for this PR as the separate commits can just be merged when the PR is accepted (and message changed at that time).

src/options-processing.cc
```diff
@@ -75,6 +75,7 @@ void service_dir_opt::build_paths(bool am_system_init)
         }
 
         done_user_home:
+        service_dir_paths.emplace_back("/etc/dinit.d/user", /*dyn_allocd=*/false);
```
davmac314, line 78: Adding the directory here unconditionally then has a problem in case the checks above failed (eg `getenv("XDG_CONFIG_HOME")` and `get_user_home()` both returned `nullptr`). The current behaviour - which perhaps isn't ideal - is to fall back on using the system manager paths. By adding the new user path here without any sort of guard, in that case _both_ `/etc/dinit.d/user` and the system paths (`/etc/dinit.d` etc as per below) will be added to the service directory path list, which I think is undesirable.

At least adding the user dirs should be guarded with a check for `home_service_dir_set`.

davmac314 (changes requested):
Minor changes requested, as noted. Specifically:

- add `/usr/local/lib/dinit.d/user` and `/usr/lib/dinit.d/user` also
- avoid mixing user and system manager paths in case finding user home fails

davmac314:
I've now merged your changes. If you'd like to add your name to the CONTRIBUTORS file, please open another PR to do that. Thanks for your contribution!

PR #323, opened 2024-04-30
meson.build: add poweroff bin

konimex:
## Description

Self-explanatory.

This will make the meson build 1:1 with the Makefile build, since the Makefile build already has poweroff but not meson.


davmac314:
Looks like the failure on MacOS is a system/infrastructure issue. @mobin-2008 maybe should install meson via `brew` rather than `pip`, if that's possible?

davmac314:
@konimex Thanks, this is merged. You should feel free to open another PR to add yourself to the CONTRIBUTORS file.

PR #326, opened 2024-05-02
src/meson.build: install poweroff, reboot, and halt as symlinks

### LINKED ISSUE #325: Meson build builds "shutdown" four times
It seems the meson build actually compiles and links the shutdown utility four times, for each of the different names:

- shutdown
- halt
- reboot
- poweroff

(the last one was added recently in a0b6a56).

These should all be links to a single binary.

konimex:
## Description

Self-explanatory.

This should close #325, unfortunately since it uses `install_symlink`, it requires a newer version of meson.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,7 +5,7 @@ project(
     version : run_command('grep', '^VERSION=', 'build/version.conf',
         check: true).stdout().strip().split('=')[1],
     license : 'Apache-2.0',
-    meson_version : '>= 0.56.0',
+    meson_version : '>= 0.61.0',
     default_options : [
         'cpp_std=c++11',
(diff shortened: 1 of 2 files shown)


mobin-2008 (approved):
Thanks.

Looks good but looks like there is no other option for keeping compatibility with older meson. I personally have a rule for required meson version:
It should be Ok for building on latest stable debian.

Debian bookworm has meson 1.0 so It's satisfied.

I will upgrade meson CI to debian bookworm.

`Reviewed-by: Mobin Aydinfar <mobin@mobintestserver.ir>`

mobin-2008:
@davmac314 Good to merge

PR #328, opened 2024-05-03
meson: Some fixes around shutdown building

mobin-2008:
Commits are self-explanatory.

There is a fix around shutdown executables and a small fix about using values from root meson.build in addition of man-pages fixes.


mobin-2008:
Hmm, looks like I pushed my branch into dinit repo. I will remove it when it got merged :/

davmac314 (approved):

PR #333, opened 2024-05-18
shutdown: Respect "SHUTDOWN_PREFIX" for reboot util check

mobin-2008:
It was a bug that shutdown executable only checked "reboot" string to find out this should be a reboot or normal shutdown.

Regression: It's not, because it was broken from the beginning.


davmac314 (approved):
Looks good, thanks

PR #334, opened 2024-05-18
dinit-log: Add text color

TheNofis:
I just added a couple of lines so that the logging was colored :D


davmac314:
Thanks for opening a PR. To accept this I would require a few changes: since the color sequences are not universal to every terminal, and since some people wouldn't want them anyway, there should be a (documented) way to turn them off (i.e. a command-line argument). Additionally I would like the escape sequences to named macros or constants rather than embed them directly in the message strings.

build/build.ninja
```diff
@@ -0,0 +1,238 @@
+# This is the build file for project "dinit"
```
mobin-2008, line 1: Looks like you pushed everything in your git working dir.

src/includes/service-constants.h
```diff
@@ -5,6 +5,12 @@
 
 #include <control-datatypes.h>
 
+// Service set log color:
+#define COLOR_GREEN "\x1b[32m" // green
+#define COLOR_RED "\x1b[33m" // red
+#define COLOR_YELLOW "\x1b[31m" // yellow
+#define COLOR_RESET "\x1b[0m" // reset
```
mobin-2008, line 12: "// green" comment and similar comments are unnecessary, Names are self-explanatory.

davmac314:
Other than cleaning up the PR so that it doesn't include a bunch of built files (as Mobin pointed out):

- I notice that the same color is used for "FAILED" as "STOPPD" status. I don't think this should be the case. A service that is "stopped" successfully does not indicate an error.
- the new `--color` option looks like it is the default anyway so the option has no effect?
- new option needs documenting in the manual pages

davmac314:
Looks like this has gone stale, so I'll close for now; feel free to re-open in future.

christmaswreath:
hey, i'm just bumping this as i love the idea of pretty colors in init and would love to see this be a thing.

Issue #291, opened 2024-02-03
Escaping a newline in dinit service causes an error


squeaktoy:
I tried to write a service with a long command so I naturally thought "let's escape the newline with backslash!" but nope, doing so results in this error:
```
Checking service: slskd...
Service 'slskd' (line 2): backslash escape (`\') not followed by character
Unable to load service 'slskd': badly formed line.
One or more errors/warnings issued.
```

The slskd service file looks like this:
```
type = process
command = bwrap --bind $HOME/.local/share/slskd $HOME/.local/share/slskd \
--bind /media/tank/archive /media/tank/archive \
--bind /media/tank/music /media/tank/music \
--bind /media/tank/movies /media/tank/movies \
--bind /media/tank/sound /media/tank/sound \
--bind /media/tank/games /media/tank/games \
--bind /media/tank/downloads/soulseek /media/tank/downloads/soulseek \
--unshare-cgroup --unshare-ipc --unshare-pid --unshare-uts \
--proc /proc \
--dev /dev \
--tmpfs /tmp \
--ro-bind /lib64 /lib64 \
--ro-bind /usr/lib64 /usr/lib64 \
--ro-bind /usr/lib /usr/lib \
--ro-bind /etc/ld.so.cache /etc/ld.so.cache \
--ro-bind /usr/local/libexec/slskd /usr/local/libexec/slskd \
--ro-bind /usr/share/zoneinfo /usr/share/zoneinfo \
--chdir /usr/local/libexec/slskd \
/usr/local/libexec/slskd/slskd
log-type = buffer
```

Seems weird to my why dinit wouldn't support this. I don't want to write a super huge command on one line.

davmac314:
> Seems weird to my why dinit wouldn't support this. I don't want to write a super huge command on one line.

This just hasn't come up before. I agree there's no real reason not to support it. In the meantime, using a shell script to start the service (via `exec`) is probably a usable workaround.

PR #345, opened 2024-06-17
Add newline escaping support

capezotte:
A few alternative options are possible (such appending the newline literally and handling them in `read_setting_value` , or translating it into a space) but this is likely what most people, coming from the shell, expect.

Closes: https://github.com/davmac314/dinit/issues/291

### DIFF
--- a/doc/manpages/dinit-service.5.m4
+++ b/doc/manpages/dinit-service.5.m4
@@ -133,7 +133,8 @@ The quote characters are not considered part of the property value.
 White space appearing inside quotes does not act as a delimiter for tokens.
 .IP \(bu
 A backslash (\\) can be used to escape the next character, causing it to
-lose any special meaning and become part of the property value.
+lose any special meaning and become part of the property value (escaped newlines are an
+exception; they mark the end of a comment, and are treated as a unescaped space).
 A double backslash (\\\\) is collapsed to a single backslash within the parameter value.
 White space preceded by a backslash will not separate tokens.
 .LP
--- a/src/includes/load-service.h
+++ b/src/includes/load-service.h
@@ -758,6 +758,7 @@ void process_service_file(string name, std::istream &service_file, T func)
     while (getline(service_file, line)) {
         unsigned line_num_after = ++line_num;
 
+        if (line.empty()) continue;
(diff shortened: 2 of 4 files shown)


davmac314:
I understand what you're saying about people coming from the shell, but that "it works the same as the shell" is not an impression we want to give at all. Could you make it an error if first character of the next line is not whitespace, since that eliminates the otherwise ambiguous case? I.e.

```
setting1 = abc\
def   # error, new line following backslash does not begin with whitespace character
setting2 = abc\
 def   # ok, setting value is "abc def"
```

A case that isn't handled particularly well by this approach is comments. Eg:
```
setting1 = something # a comment\
 comment or setting value continues here maybe
```
I think the current patch sees it as one setting with "comment or setting value continues here" being considered an extension of the comment rather than of the value. I'd prefer it was either an error or an extension of the setting value, but I'm ok with letting this through as-is for now even though technically this is a breaking change - a comment with backslash at the end was previously just a single-line comment.

davmac314:
(Otherwise, this basically looks fine to me)

capezotte:
> Could you make it an error if first character of the next line is not whitespace, since that eliminates the otherwise ambiguous case?

Done.

> I think the current patch sees it as one setting with "comment or setting value continues here" being considered an extension of the comment rather than of the value. I'd prefer it was either an error or an extension of the setting value

Now that you mentioned it, I believe it's better for it to be an extension of the setting value; in long command lines, some options (or chain-loading mini-utilities if the user goes that route) might be worth commenting on — the test file was updated accordingly.

Speaking of tests, I'd like to add tests for error conditions, but I'm not sure of the best approach; it seems each part of loadtests.cc is doing its own thing.

davmac314:
Looks good.

The "test_path_env_subst" test is probably a good template for new tests. To test for errors you can replace the exception handling there:
```
            catch (service_description_exc &exc) {
                //report_service_description_exc(exc);
            }
```
... with a version which sets a flag, that you can check at the end of the test, for example.

You can also add yourself to CONTRIBUTORS if you like.

capezotte:
That should be it.

davmac314:
Merged, thanks!

PR #353, opened 2024-07-12
Remove references to my legal name and add mailmap

q66:


davmac314:
@mobin-2008 you don't need to request review on PRs, I get a notification anyway.

PR #355, opened 2024-07-24
CI: Fix grammar and mark MacOS workflow as ARM64 not x86_64

mobin-2008:
Looks like GitHub now uses arm64 machines for MacOS workflows.

### DIFF
--- a/.github/workflows/regular_ci.yml
+++ b/.github/workflows/regular_ci.yml
@@ -82,13 +82,13 @@ jobs:
       fail-fast: false # Upload src/igr-tests/*/output/* files in igr-tests
       matrix:
         include:
-          - arch: 'amd64'
+          - arch: 'aarch64'
 
     steps:
     - uses: actions/checkout@v4
     # on MacOS, g++ symlinked to clang++
     # We must verify that
-    - name: Verify g++ (it is g++ or clang++?)
+    - name: Verify g++ (is it g++ or clang++?)
       run: g++ --version
     - name: Print g++ architecture
       run: g++ -dumpmachine


.github/workflows/regular_ci.yml
```diff
@@ -82,13 +82,13 @@ jobs:
       fail-fast: false # Upload src/igr-tests/*/output/* files in igr-tests
       matrix:
         include:
-          - arch: 'amd64'
+          - arch: 'arm64'
```
davmac314, line 85: Is there even any point listing the architecture in a matrix, if there's only one and we don't control it?

mobin-2008: Hmm, Actually I did this to make the CI list look neat (and that's should be `aarch64` not `arm64` for this purpose).

mobin-2008: Do you think it should go?

davmac314: Is there any guarantee that it runs on a particular architecture? Is it documented anywhere?

mobin-2008: No, Most of them are so implicit. Also GitHub news is not a good source for checking these because it wasn't say that macos-latest will run on arm64. Also there is README for x86_64 builds of macos in [here](https://github.com/actions/runner-images/tree/main/images/macos) but looks like they aren't available in GitHub hosted runner because [The Document](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories) says: M1 is used for MacOS-14 workflows.

davmac314: Oh so it is documented to run on M1 (that's ARM). Ok, put it as `aarch64` then.

davmac314 (approved):

PR #356, opened 2024-07-24
cptests/Makefile: Don't rebuild test objects that were built by primary tests

mobin-2008:
We already have `parent_objs` built in `../` directory with exactly same compiler flags. So why we don't use them in cptests also instead of building them again?

This also should improve build time.

### DIFF
--- a/src/tests/cptests/Makefile
+++ b/src/tests/cptests/Makefile
@@ -14,13 +14,10 @@ build-tests: cptests
 run-tests: cptests
 	./cptests
 
-cptests: cptests.o $(parent_objs) $(parent_test_objs)
-	$(CXX) -o cptests cptests.o $(parent_test_objects) $(parent_objs) $(ALL_TEST_LDFLAGS)
+cptests: cptests.o $(parent_objs:%=../%) $(parent_test_objects)
+	$(CXX) -o cptests cptests.o $(parent_test_objects) $(parent_objs:%=../%) $(ALL_TEST_LDFLAGS)
 
 $(objects): %.o: %.cc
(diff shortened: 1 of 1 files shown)


src/tests/cptests/Makefile
```diff
@@ -14,13 +14,13 @@ build-tests: cptests
 run-tests: cptests
 	./cptests
 
-cptests: cptests.o $(parent_objs) $(parent_test_objs)
-	$(CXX) -o cptests cptests.o $(parent_test_objects) $(parent_objs) $(ALL_TEST_LDFLAGS)
+# Use objects from tests/ (../) directory because we already have them
+cptests_objs = $(foreach obj,$(parent_objs),../$(obj))
```
davmac314, line 18: Creating a new variable, and using `foreach`, seems like overkill, just use a substitution reference. `$(parent_objs:%=../%)` should work and no need for a new `cptests_objs` variable.

src/tests/cptests/Makefile
```diff
@@ -14,13 +14,13 @@ build-tests: cptests
...
 
-$(objects): %.o: %.cc
-	$(CXX) $(ALL_TEST_CXXFLAGS) -MMD -MP -I../test-includes -I../../../dasynq/include -I../../../build/includes -I../../includes -c $< -o $@
+cptests: cptests.o $(cptests_objs) $(parent_test_objs)
```
davmac314, line 20: `$(parent_test_objs)` should be `$(parent_test_objects)` (it's a pre-existing mistake, but as you are changing the line then may as well fix it)

davmac314 (changes requested):

davmac314:
Also could you remove the "waste resources" from the commit message, just say what the change is without passing judgement. "cptests: don't rebuild test objects that were built by primary tests" would be fine.

davmac314 (approved):

PR #357, opened 2024-07-26
add soft-reboot

YukariChiba:
this commit adds initial support for `soft-reboot` command to `shutdown`, which is similar to [systemd's soft-reboot](https://www.freedesktop.org/software/systemd/man/latest/systemd-soft-reboot.service.html), but without root switching.

Early services need to be idempotent for soft-reboot to work properly.

---

~~Updated: added an argument `--ignore-sys-args` to ignore kernel args, which would be added automatically in soft-reboot `execv`.~~

~~Updated: filtered out "single" argument~~

Updated: The re-`execv`-ing retains boot-time parameters to simplify the code and is consistent with systemd's soft-reboot behavior.

### DIFF
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -192,7 +192,9 @@ namespace {
 
     // These need to be at namespace scope to prevent causing stack allocations when using them:
     constexpr auto shutdown_exec = literal(SBINDIR) + "/" + SHUTDOWN_PREFIX + "shutdown";
+    constexpr auto dinit_exec = literal(SBINDIR) + "/" + "dinit";
     constexpr auto error_exec_sd = literal("Error executing ") + shutdown_exec + ": ";
+    constexpr auto error_exec_dinit = literal("Error executing ") + dinit_exec + ": ";
 }
 
 // Options handled in dinit_main
@@ -684,6 +686,9 @@ int dinit_main(int argc, char **argv)
         if (shutdown_type == shutdown_type_t::REBOOT) {
             log_msg_end(" Will reboot.");
         }
+        if (shutdown_type == shutdown_type_t::SOFTREBOOT) {
+            log_msg_end(" Will soft-reboot.");
+        }
         else if (shutdown_type == shutdown_type_t::HALT) {
             log_msg_end(" Will halt.");
         }
@@ -696,6 +701,17 @@ int dinit_main(int argc, char **argv)
     close_control_socket();
     
     if (am_system_mgr) {
+        if (shutdown_type == shutdown_type_t::SOFTREBOOT) {
+            sync(); // Sync to minimise data loss if user elects to power off / hard reset
+
+            execv(dinit_exec, argv);
+            log(loglevel_t::ERROR, error_exec_dinit, strerror(errno));
+
+            // reboot if soft-reboot fails or memory allocation failed
+            log(loglevel_t::ERROR, "Could not soft-reboot. Will attempt reboot.");
+            shutdown_type = shutdown_type_t::REBOOT;
+        }
+
         if (shutdown_type == shutdown_type_t::NONE) {
             // Services all stopped but there was no shutdown issued. Inform user, wait for ack, and
             // re-start boot sequence.
--- a/BUILD_MESON
+++ b/BUILD_MESON
@@ -53,7 +53,7 @@ This command configures a "build directory" in "builddir/". You can choose anoth
 desired, but don't use a directory that is part of the Dinit source tree such as build/ or src/.
 
 The "meson setup" command will prepare to build Dinit including the dinit daemon, dinitctl utility
-and the shutdown/reboot/halt utilities by default. Tests will not be built. If you wish to change
+and the shutdown/reboot/soft-reboot/halt utilities by default. Tests will not be built. If you wish to change
 the build options, see the next section. 
 
 Once "meson setup" has completed successfully, you can compile Dinit with:
--- a/configure
+++ b/configure
@@ -143,9 +143,9 @@ Installation directories:
 Optional options:
   --enable-strip                    Strip debug information in installation process [Enabled]
   --disable-strip                   Don't strip debug information in installation process
-  --shutdown-prefix=PREFIX          Name prefix for shutdown, poweroff, reboot, halt programs []
-  --enable-shutdown                 Build shutdown, poweroff, reboot, halt programs [Enabled only on Linux systems]
-  --disable-shutdown                Don't build shutdown, poweroff, reboot, halt programs
+  --shutdown-prefix=PREFIX          Name prefix for shutdown, poweroff, reboot, soft-reboot, halt programs []
+  --enable-shutdown                 Build shutdown, poweroff, reboot, soft-reboot, halt programs [Enabled only on Linux systems]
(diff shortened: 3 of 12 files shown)


src/dinit.cc
```diff
@@ -696,6 +700,15 @@ int dinit_main(int argc, char **argv)
     close_control_socket();
     
     if (am_system_mgr) {
+        if (shutdown_type == shutdown_type_t::SOFTREBOOT) {
+            sync(); // Sync to minimise data loss if user elects to power off / hard reset
+            execv(dinit_exec, argv);
```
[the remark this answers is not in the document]
davmac314: (I just noticed this PR is still marked as WIP. Apologies if you hadn't finished yet - I will wait until you remove the WIP status before I review any further, but in general it looks ok otherwise).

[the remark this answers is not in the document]
YukariChiba: I thought of this and have several possible ideas.

1. strip the `single` argument from the `argv`, which works, but the strip process is complicated and requires more modification if new acceptable kernel arguments are added subsequently.
2. set a new cmdline argument `--no-process-sys-args`, and place it at the beginning of the `argv` when re-`execv`-ing.
3. add a new argument to `shutdown` that overrides the `argv` when re-`execv`-ing.

I think this needs further discussion.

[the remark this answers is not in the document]
YukariChiba: Added an cmdline arg `--ignore-sys-args` to ignore kernel args like `single`

src/dinit.cc
```diff
@@ -696,6 +705,30 @@ int dinit_main(int argc, char **argv)
     close_control_socket();
     
     if (am_system_mgr) {
+        if (shutdown_type == shutdown_type_t::SOFTREBOOT) {
+            sync(); // Sync to minimise data loss if user elects to power off / hard reset
+            if (!opts.ignore_sys_args) {
+                char **argv_exec = (char **)malloc((argc + 2) * sizeof(char *));
```
mobin-2008, line 711: One of the main principles of Dinit is ["Robustness (including allocation robustness)"](https://github.com/davmac314/dinit/blob/0b5158328427778f1e7d28104bc8cca43aef572a/doc/COMPARISON#L27-L31). The `malloc()` call could fail and it's not handled here and can result in null dereference in further usage (see L713).

I think malloc failure should be handled here.

davmac314: I need a little time to consider the whole issue. I think it is probably better to filter arguments out before re-exec, or even to re-build a new command line from scratch, than it is to have a new argument which effectively filters out arguments at a different stage. Please allow me a little time to get back to this issue with a response once I've thought it through. @mobin-2008 is correct though that we cannot ignore `malloc` failures. It's also generally preferable to use `new` rather than `malloc`, note that this throws an exception on failure and that needs to be handled.

YukariChiba: Okay, I've now filtered the `single` argument in `argv`, and handled the exception in `new`.

src/dinit.cc
```diff
@@ -696,6 +700,33 @@ int dinit_main(int argc, char **argv)
...
+                    if (strcmp(argv[i], "single") != 0) {
+                        argv_exec[j] = argv[i];
+                        ++j;
+                    }
+                }
+                argv_exec[j] = nullptr;
+
+                execv(dinit_exec, argv_exec);
+
+                free(argv_exec);
```
mobin-2008, line 720: As far as I know, a object created with `new[]` should be freed with `delete[]` not `free()`.
```suggestion
                delete[] argv_exec;
```

src/dinit.cc
```diff
@@ -696,6 +700,33 @@ int dinit_main(int argc, char **argv)
...
+            } catch (std::bad_alloc &exc) {
```
mobin-2008, line 721: From [Dinit's code style guide](https://github.com/davmac314/dinit/blob/master/doc/CODE-STYLE):
> 2. Curly braces - for classes and functions, the opening brace appears by itself on an empty line, as does the closing brace. In other cases the opening brace should appear at the end of the line (and not on a line by itself). The closing brace always appears on a line by itself.
```suggestion
            }
            catch (std::bad_alloc &exc) {
```

src/dinit.cc
```diff
@@ -696,6 +700,34 @@ int dinit_main(int argc, char **argv)
...
     if (am_system_mgr) {
+        if (shutdown_type == shutdown_type_t::SOFTREBOOT) {
+            sync(); // Sync to minimise data loss if user elects to power off / hard reset
+
+            try {
+                char **argv_exec = new char*[argc + 1];
+                int j = 0;
+                for (int i = 0; i < argc; i++) {
+                    // strip out kernel cmdline: single
+                    if (strcmp(argv[i], "single") != 0) {
```
davmac314, line 711: Sorry for not being able to look at this earlier. As I mentioned, I needed to find a time when I wasn't busy to figure out the right solution. Having thought about it for a bit, I think I have some better idea now.

This current solution, filtering just the "single" argument, isn't sufficient because there are a number of ways users can pass additional service names on the command line argument (via the bootloader):

- `single` (on Linux) which is covered by your current changes
- `-t servicename` for any service name or `--service servicename` for any service name
- `-m servicename1 servicename2 servicename...` for any service names
- `-c servicename1 servicename2 servicename...` for any service names (container mode)

I think the best option is to, when processing the command line arguments initially, construct an array (probably a `vector<boolean>` actually) which keeps track of which arguments should be filtered out on re-exec. Then we don't need to duplicate the argument processing logic; it's all in one place.

src/dinit.cc
```diff
@@ -696,6 +700,34 @@ int dinit_main(int argc, char **argv)
...
+                        argv_exec[j] = argv[i];
+                        ++j;
+                    }
+                }
+                argv_exec[j] = nullptr;
+
+                execv(dinit_exec, argv_exec);
```
davmac314, line 718: If the `execv` actually fails, it would be good to log why (making use of `strerror` to convert the errno into a message).

davmac314 (changes requested):

src/dinit.cc
davmac314: Alternatively, update the documentation to be very clear about what a "soft reboot" actually does. It seems reasonable to me that it will just restart with the exact same arguments, but in that case the documentation needs to be clear and warn that the original boot-time options will be re-applied.

YukariChiba: I just tested this on systemd and found that the `single` parameter is still preserved after soft-reboot, so let's just use `argv` directly to be consistent with it in this behavior.

YukariChiba: > Usecase: this is an extremely quick way to reset userspace fully when updating image based systems, without going through a full hardware/firmware/boot loader/kernel reset. It minimizes "grayout time" for OS updates. (In particular when combined with kernel live patching)

From the original PR for systemd: https://github.com/systemd/systemd/pull/27435

davmac314 (changes requested):
Otherwise looks mostly ok, although you haven't adjusted the documentation to make the precise action of a soft-shutdown more explicitly documented. I can do that myself after merging.

YukariChiba:
The behavior of soft-reboot has been described in the documentation.

davmac314 (approved):

davmac314:
Thanks, now merged. If you would like to add yourself to the CONTRIBUTORS file, please open another pull request to do so. :)

PR #364, opened 2024-08-11
dinitctl: Handle the "SERVICEEVENT vs SERVICESTATUS" race 

### LINKED ISSUE #363: dinitctl: `dinitctl enable` can get stuck in service activation status report because of missing `SERVICEEVENT`
**Describe the bug**
This is a really weird one. Sometimes the `dinitctl enable` get stuck after the "enabled" message.

**To Reproduce**
1. Have a minimal service set (including `boot` and `boot.d`)
2. Add a fail-fast service (for example a service with non-exist executable path).
3. Create a loop with disable and enable:
  ```
  #!/bin/sh
  
  SERVICE="srv"
  SOCKET="/tmp/dinitctl"
  
  i=0
  while :; do
      i=$(( i + 1 ))
      echo $i
      dinitctl disable "$SERVICE" --socket-path "$SOCKET"
      dinitctl enable "$SERVICE" --socket-path "$SOCKET"
  done
  ```
4. It will stuck after a random number of tries.

**Expected behavior**
It should not stuck.

**Additional context**
The dinitctl enables a service with this process:
1. dinitctl loads a service
2. dinitctl finds a `waits-for.d` and add the service in that directory
3. dinitctl starts the service
4. dinitctl checks the current state of service with `SERVICESTATUS` command
5. dinit responds.
6. if the status of service wasn't something that dinitctl expects, run `wait_service_state()`
7. Report the success or failure.

The problem is dinitctl expects a `SERVICEEVENT` but sometimes it's missing and dinitctl will stuck on `read()` syscall (`int r = rbuffer.fill_to(socknum, 2);`).

mobin-2008:
With this change we can eliminate the "SERVICESTATUS vs SERVICEEVENT" race.

Fixes #363

### DIFF
--- a/src/dinitctl.cc
+++ b/src/dinitctl.cc
@@ -760,27 +760,104 @@ static void print_termination_details(int exit_status)
     }
 }
 
+// Process a SERVICEEVENT packet if it is related to the specified service handle, and optionally
+// report the service status to the user (verbose == true). The caller must ensure that a complete
+// packet of type SERVICEEVENT is present in the buffer before calling.
+// The size of current SERVICEEVENT should be provided as pktlen.
+// Returns 0 if the service started (do_stop == false) or stopped (do_stop == true), 1 if
+// start/stop was cancelled or failed, -1 when the service event is not related to given service
+// handle.
+static int process_service_event(cpbuffer_t &rbuffer, unsigned pktlen, handle_t handle,
+        const std::string &service_name, bool do_stop, bool verbose)
+{
+    using std::cout;
+
+    // earlier versions do not include status info, the size in that case is
+    // base_pkt_size:
+    constexpr unsigned base_pkt_size = 2 + sizeof(handle_t) + 1;
+    if (pktlen < base_pkt_size) {
+        throw dinit_protocol_error();
+    }
+
+    service_event_t completion_event;
+    service_event_t cancelled_event;
+    if (do_stop) {
+        completion_event = service_event_t::STOPPED;
+        cancelled_event = service_event_t::STOPCANCELLED;
+    }
+    else {
+        completion_event = service_event_t::STARTED;
+        cancelled_event = service_event_t::STARTCANCELLED;
+    }
+
+    handle_t ev_handle;
+    rbuffer.extract((char *)&ev_handle, 2, sizeof(ev_handle));
+    service_event_t event = static_cast<service_event_t>(rbuffer[2 + sizeof(ev_handle)]);
+    if (ev_handle == handle) {
+        if (event == completion_event) {
+            if (verbose) {
+                cout << "Service '" << service_name << "' " << describe_state(do_stop)
+                        << ".\n";
+            }
+            rbuffer.consume(pktlen);
+            return 0;
+        }
+        else if (event == cancelled_event) {
+            if (verbose) {
+                cout << "Service '" << service_name << "' " << describe_verb(do_stop)
+                        << " cancelled.\n";
+            }
+            rbuffer.consume(pktlen);
+            return 1;
+        }
+        else if (!do_stop && event == service_event_t::FAILEDSTART) {
+            if (verbose) {
+                cout << "Service '" << service_name << "' failed to start.\n";
+                if (pktlen >= base_pkt_size + STATUS_BUFFER_SIZE) {
+                    stopped_reason_t stop_reason = static_cast<stopped_reason_t>(rbuffer[base_pkt_size + 3]);
+                    int exit_status;
+                    rbuffer.extract((char *)&exit_status, base_pkt_size + 6, sizeof(exit_status));
+
+                    switch (stop_reason) {
+                        case stopped_reason_t::DEPFAILED:
+                            cout << "Reason: a dependency of the service failed to start. "
+                                    "Check dinit log.\n";
+                            break;
+                        case stopped_reason_t::TIMEDOUT:
+                            cout << "Reason: start timed out.\n";
+                            break;
+                        case stopped_reason_t::EXECFAILED:
+                            cout << "Reason: execution of service process failed:\n";
+                            uint16_t launch_stage;
+                            rbuffer.extract((char *)&launch_stage, base_pkt_size + 4,
+                                            sizeof(uint16_t));
+                            cout << "        Stage: " << exec_stage_descriptions[launch_stage]
+                                << "\n";
+                            cout << "        Error: " << strerror(exit_status) << "\n";
+                            break;
+                        case stopped_reason_t::FAILED:
+                            cout << "Reason: service process terminated before ready: ";
+                            print_termination_details(exit_status);
+                            cout << "\n";
+                            break;
+                        default:
+                            cout << "Reason unknown/unrecognised. Check dinit log.\n";
+                    }
+                }
+            }
+            rbuffer.consume(pktlen);
+            return 1;
+        }
+    }
+    rbuffer.consume(pktlen);
+    return -1;
+}
+
 // Wait for a service to reached stopped (do_stop == true) or started (do_stop == false) state.
 // Returns 0 if the service started/stopped, 1 if start/stop was cancelled or failed.
 static int wait_service_state(int socknum, cpbuffer_t &rbuffer, handle_t handle,
         const std::string &service_name, bool do_stop, bool verbose)
 {
-    using std::cout;
-    using std::cerr;
-    using std::endl;
-
-    service_event_t completionEvent;
-    service_event_t cancelledEvent;
-
-    if (do_stop) {
-        completionEvent = service_event_t::STOPPED;
-        cancelledEvent = service_event_t::STOPCANCELLED;
-    }
-    else {
-        completionEvent = service_event_t::STARTED;
-        cancelledEvent = service_event_t::STARTCANCELLED;
-    }
-
     // Wait until service started:
     int r = rbuffer.fill_to(socknum, 2);
     while (r > 0) {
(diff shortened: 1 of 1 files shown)


src/dinitctl.cc
```diff
@@ -1822,12 +1822,44 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+            if (from_state != service_state_t::STARTED) {
+                cerr << "dinitctl: warning: enabling dependency for non-started service" << endl;
+            }
+
+            // Check the service status before sending ENABLESERVICE to avoid a
+            // "SERVICESTATUS vs SERVICEEVENT" race condition.
+            auto m = membuf()
+                    .append((char)cp_cmd::SERVICESTATUS)
+                    .append(to_handle);
+            write_all_x(socknum, m);
```
davmac314, line 1838: This introduces another race. If you check the service state before sending ENABLESERVICE, the state can change in between the check and sending ENABLESERVICE.

The best fix is probably to not use SERVICESTATUS at all, at least when enabling a service, and rely on the status information (event) packets. For disabling a service SERVICESTATUS is necessary as there is no guarantee a service will stop when it is disabled and we want to warn in case there is a separate dependency that holds it started.

davmac314 (changes requested):

src/dinitctl.cc
davmac314: > To avoid the race you would need to send some status information back in the reply to the ENABLESERVICE request.

Or, better, if possible: capture and properly hand status events that are reported between SERVICESTATUS and its reply. That might not even require changing the protocol.

src/dinitctl.cc
```diff
@@ -1903,38 +1903,43 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
-        rbuffer.consume(1);
-
-        int statussize = 6 + std::max(sizeof(pid_t), sizeof(int));;
-        fill_buffer_to(rbuffer, socknum, statussize + 1 /* reserved */);
-        rbuffer.consume(1);
-        service_state_t current = static_cast<service_state_t>(rbuffer[0]);
-        service_state_t target = static_cast<service_state_t>(rbuffer[1]);
-        rbuffer.consume(statussize);
+        catch (const dinit_protocol_error &) {
+            // We probably got a SERVICESTATUS before SERVICEEVENT, Process it.
```
davmac314, line 1920: Do not use exceptions for ordinary control flow. https://isocpp.org/wiki/faq/exceptions#why-not-exceptions

`wait_service_state` throws a protocol error if it sees an unexpected reply packet because it is not supposed to be used when expecting a reply packet, and so if it does see a reply packet then that is an error condition. Relying on the buffer still being valid and still containing the reply packet that was received is encoding an unstated assumption about the behaviour of `wait_service_state` which should not be relied on.

mobin-2008: Right. So the correct way to do that is this?
```
        fill_buffer_to(rbuffer, socknum, 1);
        if (rbuffer[0] >= 100) {
            // We got a SERVICEEVENT before the SERVICESTATUS reply.
            wait_service_state(socknum, rbuffer, to_handle, to, !enable, verbose);
        }
        else if (rbuffer[0] == (char)cp_rply::SERVICESTATUS) {
            ...
        }
        else {
            // protocol error
        }
```

davmac314: Packet ids >= 100 are reserved for information packets that can be sent at any time (out-of-band). The only type of information packet is currently `SERVICEEVENT` but that might change in the future, so it is not a good idea to assume that receiving a packet with id >= 100 means that a `SERVICEEVENT` will necessarily be sent, either.

mobin-2008: It's fixed now. I thought `ids >= 100` is OK but I just missing the second statement in `wait_service_state()` function:
https://github.com/davmac314/dinit/blob/3867cf1766134980d2c3cd6f441276217af498e9/src/dinitctl.cc#L787-L791

davmac314: Information packets that are not `SERVICEEVENT` must be correctly handled, as I mentioned:

> Packet ids >= 100 are reserved for information packets that can be sent at any time (out-of-band).

src/dinitctl.cc
```diff
@@ -1903,39 +1903,61 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+
+        if (r == -1) {
+            throw cp_read_exception(errno);
+        }
+        else if (r == 0) {
             cerr << "dinitctl: protocol error." << endl;
             return 1;
         }
-        rbuffer.consume(1);
+        else if (rbuffer[0] == (char)cp_rply::SERVICESTATUS) {
```
davmac314, line 1937: You don't need `else` clause (here or on line 1933) because if the condition is true the function exits anyway (via `return` or `throw`).

It would be good to error out early, i.e. change 

    if (rbuffer[0] == (char)cp_rply::SERVICESTATUS) {

To

    if (rbuffer[0] != (char)cp_rply::SERVICESTATUS) {

... and handle the error, then leave the other code (normal execution) outside the `if`. (That has the nice side effect that you don't end up changing the indentation of a whole bunch of code, so the diff is smaller, though that's not the primary concern).

src/dinitctl.cc
```diff
@@ -1903,39 +1903,61 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
 
-        wait_for_reply(rbuffer, socknum);
-        if (rbuffer[0] != (char)cp_rply::SERVICESTATUS) {
+        int r = rbuffer.fill_to(socknum, 2);
+        while (r > 0 && rbuffer[0] >= 100) {
+            unsigned pktlen = (unsigned char) rbuffer[1];
+            fill_buffer_to(rbuffer, socknum, pktlen);
+
+            if (rbuffer[0] == (char)cp_info::SERVICEEVENT) {
+                wait_service_state(socknum, rbuffer, to_handle, to, !enable, verbose);
```
davmac314, line 1922: Doing this for a _disable_ changes the operation. A _disable_ didn't use to wait for the service state to change, and that was intentional. (We only wait for it on _enable_ because if a user enables a service they probably want to know if it fails to start).

Also, just because a `SERVICEEVENT` was received doesn't mean that `wait_service_state` will only process a single event packet nor that it will (ever) see one of the events it is looking for. For _enable_, `wait_service_state` will wait for either `STARTED` or `STARTCANCELLED` but there is no guarantee that either will arrive before the `SERVICESTATUS` reply, and `wait_service_state` will throw a protocol error if that arrives, because is not designed to be used while waiting for a reply packet. So, you can't use it here.

Eg consider:

- ENABLESERVICE sent by dinitctl
- service is already started, ACK is queued in daemon
- service stops, SERVICEEVENT queued (with STOPPED status)
- dinitctl receives ACK
- SERVICESTATUS request sent by dinitctl
- SERVICEEVENT received by dinitctl - calls `wait_service_state` 
- SERVICESTATUS reply received, `wait_service_state` throws protocol error

There was no actual protocol error, the error was that `wait_service_state` was used when a reply packet was pending.

src/dinitctl.cc
```diff
@@ -1903,39 +1903,61 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+                return 0;
```
davmac314, line 1923: Bailing out here leaves the `SERVICESTATUS` reply still pending. In general the functions in dinitctl do not leave the communication pipe in a non-neutral state (if they return success). The reply should be consumed before returning. Not doing so could lead to bugs later if callers assume that the pipe will be in a neutral state (i.e. no replies pending).

davmac314 (changes requested):
It would be good to add a comment explaining what you are doing and why. Something like:

    For an enable, we want to wait until the service has started so we can report any 
    failure. But, if the service is already started, we won't get any service events, so we
    have to request status via SERVICESTATUS to catch that case. However, we may get
    a service event before the reply to SERVICESTATUS and in that case should use it to
    report status.

Also, the commit message is misleading. It says that wen wait for SERVICEEVENT before SERVICESTATUS reply, but in fact we wait for whichever one actually comes first.

src/dinitctl.cc
mobin-2008: The new `process_service_event()` is here to just process a service event rather than waiting for service events (unlike the `wait_service_state()`). now `wait_service_state()` and `enable_disable_service()` uses this new function.

src/dinitctl.cc
```diff
@@ -1903,30 +1926,72 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+                            service_event_t::STARTCANCELLED, false /* start */, verbose);
+                    if (ret >= 0) {
+                        // Make sure communication pipe is in a neutral state.
+                        wait_for_reply(rbuffer, socknum);
+                        if (rbuffer[0] != (char)cp_rply::SERVICESTATUS) {
+                            cerr << "dinitctl: protocol error." << endl;
+                            return 1;
+                        }
+                        fill_buffer_to(rbuffer, socknum, statussize + 1 /* reserved */);
+                        rbuffer.consume(2 + statussize);
```
davmac314, line 1963: This consumes more than it potentially filled to, that can't be right. There should never be a call to `consume` with an amount that might be more than what is in the buffer, and, the call to `fill_buffer_to` needs to make sure that the entire packet has been read into the buffer.

It should be:

                        fill_buffer_to(rbuffer, socknum, 2 + statussize);
                        rbuffer.consume(2 + statussize);

I.e. same size in both cases, to read and consume a full packet.

davmac314 (changes requested):
It's pretty good now but there is a logic error to fix and the comments could be improved I think, see individual comments.

src/dinitctl.cc
davmac314: `fill_buffer_to` is documented:

    // Fill a circular buffer from a file descriptor, until it contains at least _rlength_ bytes.

That's _not_ in addition to the number of bytes already in the buffer - it's the total number of bytes. When `wait_for_reply` calls it with an argument of 1, it ensures there is at least 1 byte in the buffer. That doesn't affect later calls.

If you then call it with `statussize + 1` then there are least `statussize + 1` bytes in the buffer when it returns, and possibly _exactly_ that amount. That's not a full SERVICESTATUS packet, it's one byte short.

davmac314:
I'm unsubscribing from notifications, please request review when ready.

src/dinitctl.cc
```diff
@@ -1903,30 +1929,68 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+                    int ret = process_service_event(rbuffer, pktlen, to_handle, to, false /* start */,
+                            verbose);
+                    if (ret >= 0) {
+                        // Consume the outstanding reply packet before returning.
+                        wait_for_reply(rbuffer, socknum);
+                        if (rbuffer[0] != (char)cp_rply::SERVICESTATUS) {
+                            cerr << "dinitctl: protocol error." << endl;
+                            return 1;
+                        }
+                        fill_buffer_to(rbuffer, socknum, statussize + 2 /* 1 byte reserved */);
```
davmac314, line 1965: The comment here: `/* 1 byte reserved */` - doesn't explain the `+ 2` in the code. Something like: `/* 1 byte packet type, 1 byte reserved */` would make more sense.

src/dinitctl.cc
```diff
@@ -760,6 +760,81 @@ static void print_termination_details(int exit_status)
...
+                    default:
+                        cout << "Reason unknown/unrecognised. Check dinit log.\n";
+                    }
+                }
+            }
+            return 1;
+        }
+    }
+    return -1;
+}
```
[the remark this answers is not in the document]
davmac314: You need to adjust `wait_service_state` appropriately. I mentioned this in the review comment:

> (That will mean adjusting wait_service_state also).

It should not try to consume bytes that have been consumed already.

src/dinitctl.cc
```diff
@@ -1903,30 +1926,72 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+                    }
+                }
+
+                rbuffer.consume(pktlen);
+                r = rbuffer.fill_to(socknum, 2);
+            }
+            if (r == -1) {
+                throw cp_read_exception(errno);
+            }
+            else if (r == 0) {
```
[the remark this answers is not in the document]
davmac314: You removed the whole `if (r == 0)` clause instead of just removing the `else`.

src/dinitctl.cc
```diff
@@ -760,27 +760,106 @@ static void print_termination_details(int exit_status)
     }
 }
 
-// Wait for a service to reached stopped (do_stop == true) or started (do_stop == false) state.
-// Returns 0 if the service started/stopped, 1 if start/stop was cancelled or failed.
-static int wait_service_state(int socknum, cpbuffer_t &rbuffer, handle_t handle,
+// Process a SERVICEEVENT packet if it is related to the specified service handle, and optionally report
```
davmac314, line 763: See style guide, lines should wrap before 100 characters.

> Line widths - do not exceed 110 characters; try to limit to 100.

The only reason to go beyond 100 is if it is awkward to break the line beforehand or if it will naturally end just slightly over.

src/dinitctl.cc
```diff
@@ -760,27 +760,106 @@ static void print_termination_details(int exit_status)
...
+// the service status to the user (verbose == true).
```
davmac314, line 764: This line, though, is short. The next sentence shouldn't be on a new line. That isn't done anywhere else, please stick to conventions that are in use.

src/dinitctl.cc
```diff
@@ -760,27 +760,106 @@ static void print_termination_details(int exit_status)
...
+// The caller must ensure that a complete packet of type SERVICEEVENT is present in the buffer before
+// calling.
+// The size of current SERVICEEVENT should be provided as "pktlen".
```
davmac314, line 767: Here you have quotes around the parameter name (`"pktlen"`) but on the next line there are no quotes around another parameter name (`do_stop`). Please make sure changes are consistent - with existing practice, and with themselves.

src/dinitctl.cc
```diff
@@ -1903,30 +1929,68 @@ static int enable_disable_service(int socknum, cpbuffer_t &rbuffer, service_dir_
...
+        // a service event before the reply to SERVICESTATUS and in that case should use it to
+        // report status.
+        if (enable) {
+            int r = rbuffer.fill_to(socknum, 2);
+            while (r > 0 && rbuffer[0] >= 100) {
+                unsigned pktlen = (unsigned char) rbuffer[1];
+                fill_buffer_to(rbuffer, socknum, pktlen);
+
+                if (rbuffer[0] == (char)cp_info::SERVICEEVENT) {
+                    int ret = process_service_event(rbuffer, pktlen, to_handle, to, false /* start */,
```
davmac314, line 1956: wrap at 100, that means the `false` parameter moves to the next line

davmac314 (changes requested):
Please double-check previous review comments and review style yourself before requesting review - there are some things here you could easily catch yourself I think.

src/dinitctl.cc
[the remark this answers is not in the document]
mobin-2008: `wait_service_state()` and `enable_disable_service()` are fixed to not consume already consumed data. Sorry about this.

src/dinitctl.cc
mobin-2008: it will be too ugly:
```
fill_buffer_to(rbuffer, socknum,
        statussize + 2 /* 1 byte packet type, 1 byte reserved */);
```
I wrote it in a separate line.

src/dinitctl.cc
mobin-2008: My idea was a new line for each but looks like it's better to be combined.

src/dinitctl.cc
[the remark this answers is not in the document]
mobin-2008: Oops, Fixed.

mobin-2008:
@davmac314 I double checked my changes and your comments and I think it's ready for review.

davmac314 (approved):
Looks good, thanks.

PR #372, opened 2024-09-10
dinitctl: Error on empty arguments

### LINKED ISSUE #365: dinitctl: dinitctl will get stuck when a empty argument is passed as service_name
**Describe the bug**
The title is self-explanatory.

**To Reproduce**
1. Use `dinitctl start ""` or `dinitctl enable ""`.
2. Dinitctl is stuck because it pass a empty service name to Dinit daemon but it's unexpected and Dinit will assume that is a partial command and Dinit will wait forever.

**Expected behavior**
It should print a error and then exit without any contact with Dinit.

**Additional context**
I found this when I was trying to fix #363 when I accidentally used `"$SRVICE"` in my dinitctl commands.

mobin-2008:
An empty argument means that something should be there but it's not there for some reason (for example an empty shell variable is used). dinitctl should exit on empty argument.

Fixes #365 

### DIFF
--- a/src/dinitctl.cc
+++ b/src/dinitctl.cc
@@ -153,7 +153,7 @@ int dinitctl_main(int argc, char **argv)
             }
             else if (strcmp(argv[i], "--socket-path") == 0 || strcmp(argv[i], "-p") == 0) {
                 ++i;
-                if (i == argc) {
+                if (i == argc || argv[i][0] == '\0') {
                     cerr << "dinitctl: --socket-path/-p should be followed by socket path" << std::endl;
                     return 1;
                 }
@@ -165,7 +165,7 @@ int dinitctl_main(int argc, char **argv)
             else if (strcmp(argv[i], "--from") == 0) {
                 if (command == ctl_cmd::ENABLE_SERVICE || command == ctl_cmd::DISABLE_SERVICE) {
                     ++i;
-                    if (i == argc) {
+                    if (i == argc || argv[i][0] == '\0') {
                         cerr << "dinitctl: --from should be followed by a service name" << std::endl;
                         return 1;
                     }
@@ -204,7 +204,7 @@ int dinitctl_main(int argc, char **argv)
                 }
             }
             else if (strcmp(argv[i], "--services-dir") == 0 || strcmp(argv[i], "-d") == 0) {
-                if (++i < argc) {
+                if (++i < argc && argv[i][0] != '\0') {
                     service_dir_opts.set_specified_service_dir(argv[i]);
                 }
                 else {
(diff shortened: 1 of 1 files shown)


mobin-2008:
The empty argument is considered an error.

davmac314 (changes requested):
Could you move the main empty check (lines 127-131) to the command argument handling section? (around line 296, under `// service name / other non-option`).

The argument to `--services-dir` should also be checked.

Use `argv[i][0]` instead of `*argv[i]` in the checks on line 161/173 (and for `--services-dir`), be consistent with what is done elsewhere. Did you look at dinit.cc/dinitcheck.cc?

src/dinitctl.cc
```diff
@@ -294,6 +294,11 @@ int dinitctl_main(int argc, char **argv)
         }
         else {
             // service name / other non-option
+            if (argv[i][0] == '\0') {
+                // Empty argument
+                cmdline_error = true;
```
davmac314, line 299: This is ok for now, but it would have been a good opportunity to give a more specific error message for this specific error.

davmac314 (dismissed):

src/dinitctl.cc
mobin-2008: Is this OK?
```c
    if (empty_arg_error) {
        cerr << "dinitctl: Invalid empty argument in command line\n";
        cerr << "Try 'dinitctl --help' for more information and check the command line\n";
        return 1;
    }
```

davmac314: You don't need "in command line", or the 2nd line at all. Most of the errors are just a single line:

```
                        cerr << "dinitctl: --from should be followed by a service name" << std::endl;
                        return 1;
```

```
                    cerr << "dinitcheck: '--services-dir' (-d) requires an argument" << endl;
                    return 1;
```

```
                cerr << "dinitctl: unrecognized/invalid option: " << argv[i] << " (use --help for help)\n";
                return 1;
```

So be consistent with that!. You could put " (use --help for help)" at the end which would at least be consistent with that last example but I don't think that's necessary in this case, is the help really going to be helpful to the user for this error? - Think about it.

I know there is one case where it does say "Try 'dinitctl --help' for more information" - that is the generic error case where it is possible the user tried to use a command that isn't recognised for example (maybe they forgot the command or mis-spelled it. Then, seeing the help might be useful). But that is a very generic case. This is a specific error.

davmac314: So for example:

    cerr << "dinitctl: invalid empty argument\n"

... is fine.

src/dinitctl.cc
```diff
@@ -294,6 +295,10 @@ int dinitctl_main(int argc, char **argv)
         }
         else {
             // service name / other non-option
+            if (argv[i][0] == '\0') {
+                empty_arg_error = true;
```
davmac314, line 299: You can just report the error here. (What benefit is there in setting a variable here and testing it later? Look at how other errors are reported (other than the generic case), and be consistent with that).

davmac314 (changes requested):

davmac314 (approved):

PR #377, opened 2024-09-15
CI: Don't use "DEBUG=1" for igr-tests in FreeBSD workflow

mobin-2008:


davmac314:
Could you add to the commit message explaining why it's being removed? eg it has no effect anymore.

mobin-2008:
I did add a simple description to commit and eliminated all usage of `DEBUG=1` in CI workflows (Not only FreeBSD one).

davmac314 (approved):

PR #381, opened 2024-09-20
shutdown: Select respective action on halt and soft-reboot utilities

mobin-2008:
The #333 came to my mind and I checked the `shutdown.cc` code to make sure that's OK and It's not.

### DIFF
--- a/src/shutdown.cc
+++ b/src/shutdown.cc
@@ -33,6 +33,7 @@ static constexpr uint16_t min_cp_version = 1;
 static constexpr uint16_t max_cp_version = 1;
 
 static constexpr auto reboot_execname = cts::literal(SHUTDOWN_PREFIX) + cts::literal("reboot");
+static constexpr auto soft_reboot_execname = cts::literal(SHUTDOWN_PREFIX) + cts::literal("soft-reboot");
 
 using loop_t = dasynq::event_loop_n;
 using rearm = dasynq::rearm;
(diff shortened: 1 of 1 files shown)


src/shutdown.cc
```diff
@@ -257,9 +259,15 @@ int main(int argc, char **argv)
     auto shutdown_type = shutdown_type_t::POWEROFF;
 
     const char *execname = base_name(argv[0]);
-    if (strcmp(execname, reboot_execname) == 0) {
+    if (strcmp(execname, halt_execname) == 0) {
+        shutdown_type = shutdown_type_t::HALT;
+    }
+    else if (strcmp(execname, reboot_execname) == 0) {
```
davmac314, line 265: From the manual page:

       The default action is to power down the system if called as either
       "shutdown", "halt", or "poweroff"

This breaks that.

src/shutdown.cc
```diff
@@ -257,9 +259,15 @@ int main(int argc, char **argv)
...
         shutdown_type = shutdown_type_t::REBOOT;
     }
+    else if (strcmp(execname, soft_reboot_execname) == 0) {
+        shutdown_type = shutdown_type_t::SOFTREBOOT;
+    }
```
davmac314, line 270: This is ok but the manual page needs updating.

davmac314 (changes requested):

src/shutdown.cc
mobin-2008: It's confusing for me that `halt` and `shutdown -h` have different effect but I'm  going to fix it to match the doc.

src/shutdown.cc
mobin-2008: From shutdown.8:

> -s
> Restart the service manager and all user-space services without restarting the system.
> This is the default if executed as soft-reboot.

I don't see what should get updated in this doc.

davmac314: You're right, it's already ok. I was looking at the previous version man page by mistake.

src/shutdown.cc
mobin-2008: The halt check is removed

davmac314 (approved):

mobin-2008:
@davmac314 This is my last patch before the next release.

PR #386, opened 2024-10-02
add support for passing arguments to services

q66:
The argument is a part of the full service name always. Therefore, each argument instantiation is a separate instance of the service, with separate loading.

The following cases are supported:

1) passing the argument to any `dinitctl` command that takes an a service name
2) dependencies with an argument
2) using the argument in any context that does variable substitution
3) using the argument in dependency arguments (performing a restricted version of variable substitution)

documentation is updated, load+integration tests are updated/added

while at it, i found that depends-ms.d was being parsed wrong (it was never reading the dir) so i fixed that too

### DIFF
--- a/src/igr-tests/igr-runner.cc
+++ b/src/igr-tests/igr-runner.cc
@@ -40,6 +40,7 @@ void catlog_test();
 void offline_enable_test();
 void xdg_config_test();
 void cycles_test();
+void svc_arg_test();
 
 int main(int argc, char **argv)
 {
@@ -55,7 +56,8 @@ int main(int argc, char **argv)
             { "pseudo-cycle", pseudo_cycle_test }, { "before-after", before_after_test},
             { "before-after2", before_after2_test }, { "log-via-pipe", log_via_pipe_test },
             { "catlog", catlog_test }, { "offline-enable", offline_enable_test },
-            { "xdg-config", xdg_config_test }, { "cycles", cycles_test } };
+            { "xdg-config", xdg_config_test }, { "cycles", cycles_test },
+            { "svc-arg", svc_arg_test } };
     constexpr int num_tests = sizeof(tests) / sizeof(tests[0]);
 
     dinit_bindir = "../..";
@@ -973,4 +975,24 @@ void cycles_test()
 
     igr_assert_eq(read_file_contents(igr_input_basedir + "/cycles/expected-after_self"),
             dinitctl_p.get_stderr());
+}
+
+void svc_arg_test()
+{
+    igr_test_setup setup("svc-arg");
+    std::string output_file = setup.prep_output_file("svc-arg-record");
+    std::string socket_path = setup.prep_socket_path();
+
+    igr_env_var_setup env_output("OUTPUT", output_file.c_str());
+
+    dinit_proc dinit_p;
+    dinit_p.start("svc-arg", {"-u", "-d", "sd", "-p", socket_path, "-q", "checkarg@foo"});
+    dinit_p.wait_for_term({1,0} /* max 1 second */);
+
+    check_file_contents(output_file, std::string() +
+            "test-$1\n" +
+            "test-hello\n" +
+            "hello\n" +
+            "foo\n" +
+            "foo\n");
 }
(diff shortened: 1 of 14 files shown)


src/igr-tests/igr-runner.cc
```diff
@@ -55,7 +56,8 @@ int main(int argc, char **argv)
             { "pseudo-cycle", pseudo_cycle_test }, { "before-after", before_after_test},
             { "before-after2", before_after2_test }, { "log-via-pipe", log_via_pipe_test },
             { "catlog", catlog_test }, { "offline-enable", offline_enable_test },
-            { "xdg-config", xdg_config_test }, { "cycles", cycles_test } };
+            { "xdg-config", xdg_config_test }, { "cycles", cycles_test },
+            { "arg", arg_test } };
```
mobin-2008, line 60: The `arg` is too generic for this test, It's confusing in the first look. I suggest to use `svc-arg` instead of just `arg` for this test.

q66:
i'm a bit confused by the aarch64 failure after i force pushed?

q66:
actually that was probably just a readiness race for the dependency chain, i made all the dependencies scripted (that should address that)

davmac314:
Ok, I've looked at this fairly thoroughly. The only issue I have is that it's introducing a separate function to do another kind of substitution (what you've referred to in the docs as "minimal variable substitution"). While I get the point of minimal substitution (i.e. mostly to simplify implementation of dependency resolution) and I'm somewhat ok with not having environment variables available for these substitutions, I don't see any real point in having different rules and and especially implementation of the substitution function.

Why not make substitution for dependencies just support the full, normal substitution syntax, with the only caveat that the substitution happens before the `env-file` is loaded?

Also, I think that `$N` for any integer `N` (not just 1, not even just a single digit) should be parsed as a variable. Only `$1` may actually resolve to a service argument, of course, but it makes me uneasy that for example `$2` undergoes no substitution but `$1` does. I also think that it would be better if a reference to a missing argument gave an error, rather than substituting a blank value (so `$2` for example would always give an error, and `$1` would give an error if no argument had been supplied). I just can't see a case where an optional argument would be useful (let me know if you do have something specific in mind though), and I think that silently allowing it may raise confusion if the service doesn't work as expected as a result.

A couple of minor formatting nits: there is an `} else j++;` i.e. else on the same line as the closing brace, the existing style is to always put `else` on the next line. In man page sources, new sentences should always start on a new line (or putting it another way, ends of sentences should always be at the end of a line) - there are a few existing exceptions to that, those are actually mistakes. (For the man pages this can apparently affect the spacing of output when they are converted to printable format). I will likely go over the man pages and tweak wording though anyway, so I can fix the formatting for those up myself.

q66:
yeah that's fair, i can make these changes

wrt manpages we should probably consider moving to something better to generate them, e.g. https://git.sr.ht/~sircmpwn/scdoc because the m4 macros are a serious pain

davmac314:
You mean the roff macros, right? Yeah I've been thinking for a while about a better solution. `scdoc` doesn't look like it supports the classic `.TP`-style paragraphs though ("definition list" style) which would be a deal breaker unless there is a workaround. I may have a play around with it.

q66:
i did the changes as you asked, let me know if it's as you wanted it

src/includes/load-service.h
```diff
@@ -1215,6 +1233,19 @@ static void value_var_subst(const char *setting_name, std::string &line,
     line = std::move(r_line);
 }
 
+// Reads a dependency name while performing minimal argument expansion in it.
+inline string read_dependency_value(const char *setting_name, file_pos_ref input_pos, string_iterator &i,
+        string_iterator end, std::list<std::pair<unsigned,unsigned>> &offsets, const char *argval)
+{
+    string rval;
+    if (!argval) argval = "";
```
davmac314, line 1241: Does this prevent erroring out in case `$1` is used but no service argument has been supplied?

q66: that was a leftover from when i was doing the custom expansion and forgot to remove it, fixed now

davmac314:
Maybe a couple of other minor things, I'll do a full review this evening (AU time).

davmac314 (approved):

davmac314:
I need to clean up the loading code anyway, for now i'm merging as-is.

PR #388, opened 2024-10-03
Implement listening on activation environment changes

q66:
This implements a protocol that lets a connection to the control socket listen on changes in the global environment issued by `dinitctl setenv`. This opens up a range of possibilities - such as special services that trigger other services when an environment variable appears.

It also implements a `dinit-monitor` environment watching mode as a proof of functionality, though likely not acceptable as is (likely at least the `string_view` will need getting rid of?)

I intentionally did not implement things like protocol checking and whatnot yet since I want your feedback first.

I should probably also change the protocol a bit - at very least, the envevent should contain information about whether the envvar is newly created or if it overwrites a previous one.

Let me know what you think :)

### DIFF
--- a/src/dinit-monitor.cc
+++ b/src/dinit-monitor.cc
@@ -2,6 +2,7 @@
 #include <vector>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 
 #include <cstring>
 #include <csignal>
@@ -32,10 +33,13 @@ struct stringview {
 static std::vector<stringview> split_command(const char *cmd);
 static bool load_service(int socknum, cpbuffer_t &rbuffer, const char *name, handle_t *handle,
         service_state_t *state);
+static void request_environ(int socknum, cpbuffer_t &rbuffer, uint16_t proto_version);
+static size_t get_allenv(int socknum, cpbuffer_t &rbuffer);
 
 // dummy handler, so we can wait for children
 static void sigchld_handler(int) { }
-void issue_command(const char* service_name, const char* event_str,std::vector<stringview> &command_parts);
+static void issue_command(const char* name, const char* value, const char* event, std::vector<stringview> &command_parts, bool is_env = false);
+static size_t read_var_and_issue(int socknum, cpbuffer_t &rbuffer, size_t dsz, const std::unordered_set<std::string> &varset, std::string &enval, std::vector<stringview> &command_parts, bool &issued, const char* str_set, const char* str_unset);
 
 int dinit_monitor_main(int argc, char **argv)
 {
@@ -44,9 +48,13 @@ int dinit_monitor_main(int argc, char **argv)
     const char *control_socket_path = nullptr;
     bool user_dinit = (getuid() != 0);  // communicate with user daemon
     bool issue_init = false;  // request initial service state
+    bool use_environ = false;  // listening on activation environment changes
+    bool exit_after = false;  // exit after first issued command
     const char *str_started = "started";
     const char *str_stopped = "stopped";
     const char *str_failed = "failed";
+    const char *str_set = "set";
+    const char *str_unset = "unset";
 
     const char *command_str = nullptr;
     std::vector<const char *> services;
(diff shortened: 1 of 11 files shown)


q66:
i think the code quality in this is ok now, what's missing from my side is imo:

1) documentation
2) tests
3) the protocol bit about new/overriden envvars

q66:
i also added an option to `dinit-monitor` that makes it exit upon first issued command; that means you can invoke something like

```
$ dinit-monitor --initial --env --exit --command /usr/bin/true FOO BAR BAZ
```

as a way to wait until either of the given environment variables becomes available (with `--initial` it should be race-free and will simply exit when it's already available)

q66:
There is another thing I would like to have while at it and that is unsetting of environment vars. As I see it, we don't have to add a protocol message for it; we can repurpose the `SETENV` message and instead of raising `BADREQ` when the value has no `=`, we could just change

```
    eq = envVar.find('=');
    if (!eq || eq == envVar.npos) {
        // Not found or at the beginning of the string
        goto badreq;
    }

    main_env.set_var(std::move(envVar));
```

to

```
    eq = envVar.find('=');
    if (!eq || eq == envVar.npos) {
        // Unset
        main_env.undefine_var(std::move(envVar));
    } else {
        main_env.set_var(std::move(envVar));
    }
```

and add a new command to `dinitctl` that will invoke the protocol in this form...

thoughts?

davmac314:
> Let me know what you think :)

This can go in if you need it. I'm sort of questioning whether this is really the best option though - seems like you basically just need a channel to communicate that _something_ has happened (display server has started, or whatever), and dinit (with a little augmentation) just happens to be able to act as that channel. Could triggered services achieve the same thing? I.e. just trigger some service after setting the environment variable (possibly one service per variable, and you also have the option of having a dependent service which then will only start once all variables have been triggered, for example).

> we can repurpose the SETENV message and instead of raising BADREQ when the value has no =, we could just change

This seems perfectly reasonable to me.

q66:
well we're still using triggered services either way

the thing is stuff like `gnome-session` updates the activation environment in dbus, and dbus can in turn update the environment in dinit (which it should, especially with dbus-broker)

with this we can have dinit automatically react to it and trigger the right triggered service, without having to patch purpose-specific paths into either dbus, or every single desktop session program, so all the "specific" logic is in things that are particular to dinit services, and the only patched thing is dbus, which is patched with generic logic to update dinit when it's itself updated

we are pretty much doing the same thing as systemd in this regard, so it plays along with what everybody already does; having to patch things for dinit in different places would be much more of a pain

q66:
I implemented the rest of what I wanted to do. There is now `dinitctl unsetenv` and it works. The event protocol tracks override (hopefully correctly) and both sets and unsets (for unsets, the override flag is true when the variable previously existed).

I updated the manpages too, not sure what to do with tests considering this is only protocol and `dinit-monitor` and we don't have tests for that.

davmac314:
Ok, looks fine. For tests you can add something to `src/tests/cptests/cptests.cc`, that is where control protocol tests live.

q66:
I added a cptest, seems to be fine

src/dinit-monitor.cc
```diff
@@ -480,3 +558,95 @@ static bool load_service(int socknum, cpbuffer_t &rbuffer, const char *name, han
...
+        }
+        else {
+            issue_command(sp, &sp[eq + 1], str_set, command_parts, true);
+        }
+        return true;
+    }
+    return false;
+}
+
+static size_t read_and_issue(int socknum, cpbuffer_t &rbuffer, size_t dsz, const std::unordered_set<std::string> &varset,
```
davmac314, line 621: Could we name this `read_vars_and_issue` or something a little more specific? (I would tend to use `process_initial_vars` but whatever, just needs to be a bit more descriptive)

src/dinit-monitor.cc
```diff
@@ -480,3 +558,95 @@ static bool load_service(int socknum, cpbuffer_t &rbuffer, const char *name, han
...
+    wait_for_reply(rbuffer, socknum);
+
+    cp_rply reply_pkt_h = (cp_rply)rbuffer[0];
+    if (reply_pkt_h != cp_rply::ACK) {
+        throw dinit_protocol_error();
+    }
+    rbuffer.consume(1);
+}
+
+static size_t get_allenv(int socknum, cpbuffer_t &rbuffer)
```
davmac314, line 576: So this reads and processes the packet header and leaves the body of the reply pending? Could you add a comment explaining that

davmac314:
Sorry, just a couple more minor nits (comments above).

One more thing is that really, dinit-monitor should check for a compatible protocol version before using the `LISTENENV` packet. Or raise the minimum version that dinit-monitor will talk. In either case "server is too old" is a better message than "protocol error".

src/dinit-monitor.cc
q66: done

src/dinit-monitor.cc
q66: yes, added

q66:
i fixed the above, and also incremented the protocol ver to allow it to work at all

davmac314 (approved):

davmac314:
Great, thanks

PR #391, opened 2024-10-08
Fix status filling for new protocol

q66:
Just something I ran into while browsing the code for libdinitctl. This new protocol does not seem to be used at all yet because the protocol version has not been incremented to 5, but might as well not forget.

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -754,7 +754,7 @@ bool control_conn_t::list_services5()
             pkt_buf[0] = (char)cp_rply::SVCINFO;
             pkt_buf[1] = nameLen;
 
-            fill_status_buffer(&pkt_buf[2], sptr);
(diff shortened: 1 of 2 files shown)


q66:
i've also thrown in a free fix for accidental VLA usage (which causes a compiler warning)

davmac314:
> i've also thrown in a free fix for accidental VLA usage

That's not a VLA, the `get_size()` function is `constexpr`. I think your compiler has a bug. I can't reproduce that. See https://godbolt.org/z/7bs7vMrxo

davmac314:
> This new protocol does not seem to be used at all yet because the protocol version has not been incremented to 5

Ah damn, that was a mistake. I should've bumped the protocol version at release.

davmac314 (approved):

PR #394, opened 2024-10-13
Service argument enhancements

q66:
This changes 3 things:

1) The meta-commands now resolve the service argument. This is useful to set different fields depending on the argument value (by including separate per-arg-value file)
2) The env-file path itself performs minimal expansion (so you can include different env-files depending on the value of the argument)
3) Make it possible to have a service that only optionally takes an argument by using the substitution syntax that takes default values instead of always erroring. The syntaxes without a value will still error.


davmac314 (approved):
Thanks, looks fine

PR #395, opened 2024-10-13
Implement auto-recovery

### LINKED ISSUE #387: activatable script/top level service on failure of "boot" one
On a normal PC when the boot fails you can at least read something on the display (even if not ideal and usually the failing service is way at the top and scrolled past already at the end when all boot services are stopped) and interact with the usual [`Choose: (r)eboot, r(e)covery, re(s)tart boot sequence, (p)ower off? `](https://github.com/davmac314/dinit/blob/116633b/src/dinit.cc#L802) options; however in my use case (a random Android phone with vendor kernel) that's not a practical option due to typically non-functional VT framebuffer (not to mention the need to use a USB keyboard via OTG, so I wrote this pretty bad patch which does work wonderfully for my needs:
```diff
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -193,6 +193,9 @@ namespace {
     // These need to be at namespace scope to prevent causing stack allocations when using them:
     constexpr auto shutdown_exec = literal(SBINDIR) + "/" + SHUTDOWN_PREFIX + "shutdown";
     constexpr auto error_exec_sd = literal("Error executing ") + shutdown_exec + ": ";
+    // HYBRIS HACK: Panic command handler in case dinit fails to boot the system
+    constexpr auto panic_exec = literal(SBINDIR) + "/" + "dinit-panic";
+    constexpr auto error_exec_panic = literal("Error executing ") + panic_exec + ": ";
 }
 
 // Options handled in dinit_main
@@ -700,7 +703,17 @@ int dinit_main(int argc, char **argv)
             // Services all stopped but there was no shutdown issued. Inform user, wait for ack, and
             // re-start boot sequence.
             sync(); // Sync to minimise data loss if user elects to power off / hard reset
-            confirm_restart_boot();
+            // HYBRIS HACK: There's likely no working VT for user input in confirm_restart_boot()
+            log_msg_end(" Will run dinit-panic or reboot.");
+            flush_log();
+            sync();
+
+            execl(panic_exec.c_str(), panic_exec.c_str(), nullptr);
+            log(loglevel_t::ERROR, error_exec_panic, strerror(errno));
+            flush_log();
+            sync();
+            shutdown_type = shutdown_type_t::REBOOT;
+
             if (services->count_active_services() != 0) {
                 // Recovery service started
                 goto run_event_loop;
```

Would anything like that be interesting here for upstream? To be clear this more likely would take the form of a `boot`-adjacent "top level service" like `recovery` or maybe `panic` which can be activated via configuration around https://github.com/davmac314/dinit/blob/116633b/src/dinit.cc#L721 instead of calling the hardcoded `confirm_restart_boot()` function assuming working display/keyboard input.

For further context https://github.com/JamiKettunen/chimera-libhybris/blob/8adfd6d/overlays/base/usr/bin/dinit-panic is how I use the feature to setup a USB network between the device side and a host PC it's attached to so I can `ssh` in to inspect and fix "the damage" or whatever may have gone wrong and prevented a normal boot from finishing :)

JamiKettunen:
In some setups it makes sense to launch dinit with `--auto-recovery` and handle boot failures purely via the `recovery` service always. Likely missing some stuff etc but this implements the feature in a hopefully proper way.

Fixes #387.

### DIFF
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -79,6 +79,7 @@ static dirload_service_set *services;
 
 static bool am_system_mgr = false;     // true if we are PID 1
 static bool am_system_init = false; // true if we are the system init process
+static bool auto_recovery = false;  // automatically run recovery service on boot failure
 
 static bool did_log_boot = false;
 static bool control_socket_open = false;
@@ -292,6 +293,9 @@ static int process_commandline_arg(char **argv, int argc, int &i, options &opts)
             am_system_mgr = false;
             opts.process_sys_args = false;
         }
+        else if (strcmp(argv[i], "--auto-recovery") == 0 || strcmp(argv[i], "-r") == 0) {
+            auto_recovery = true;
+        }
         else if (strcmp(argv[i], "--socket-path") == 0 || strcmp(argv[i], "-p") == 0) {
             if (++i < argc && argv[i][0] != '\0') {
                 control_socket_path = argv[i];
@@ -391,6 +395,7 @@ static int process_commandline_arg(char **argv, int argc, int &i, options &opts)
                     " --system-mgr, -m             run as system manager (perform shutdown etc)\n"
                     " --user, -u                   run as a user service manager\n"
                     " --container, -o              run in container mode (do not manage system)\n"
+                    " --auto-recovery, -r          auto-run recovery service on system manager boot failure\n"
                     " --socket-path <path>, -p <path>\n"
                     "                              path to control socket\n"
                     " --ready-fd <fd>, -F <fd>\n"
@@ -695,6 +700,9 @@ int dinit_main(int argc, char **argv)
         else if (shutdown_type == shutdown_type_t::POWEROFF) {
             log_msg_end(" Will power down.");
         }
+        else if (shutdown_type == shutdown_type_t::NONE) {
+            log_msg_end(" Will handle boot failure.");
+        }
     }
 
     flush_log();
@@ -718,6 +726,16 @@ int dinit_main(int argc, char **argv)
             // Services all stopped but there was no shutdown issued. Inform user, wait for ack, and
             // re-start boot sequence.
             sync(); // Sync to minimise data loss if user elects to power off / hard reset
+            if (auto_recovery) {
+                try {
+                    services->start_service("recovery");
+                }
+                catch (std::exception &exc) {
+                    log(loglevel_t::ERROR, "Unable to start recovery service: ", exc.what());
+                    // As the following prompt UI could be inaccessible flush the log again already
+                    flush_log();
+                }
+            }
             confirm_restart_boot();
             if (services->count_active_services() != 0) {
                 // Recovery service started
--- a/doc/manpages/dinit.8.m4
+++ b/doc/manpages/dinit.8.m4
@@ -97,6 +97,11 @@ Run in "container mode", i.e. do not perform system management functions (such
 as shutdown/reboot).
 The \fBdinit\fR daemon will simply exit rather than executing the \fB$$$SHUTDOWN_PREFIX@@@shutdown\fR program.
 .TP
+\fB\-r\fR, \fB\-\-auto\-recovery\fR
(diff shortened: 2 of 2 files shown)


davmac314:
Thanks - I will try to review soon

src/dinit.cc
```diff
@@ -718,7 +723,19 @@ int dinit_main(int argc, char **argv)
             // Services all stopped but there was no shutdown issued. Inform user, wait for ack, and
             // re-start boot sequence.
             sync(); // Sync to minimise data loss if user elects to power off / hard reset
-            confirm_restart_boot();
+            if (auto_recovery) {
+                try {
+                    log_msg_end(" Will start recovery.");
```
davmac314, line 728: `log_msg_end` is supposed to be paired with `log_msg_begin` (and zero or more `log_msg_part`s). To issue a message that doesn't consist of multiple parts you can use a single call to `log` instead.

The message `" Will start recovery."` is not self-contained (and starts with a space). I think maybe you copied and modified from one of the lines above, which are all part of a two-part message. If you log a message, it should be self-contained and should not start with white space. I would suggest just `"Starting recovery."`.

src/dinit.cc
```diff
@@ -718,7 +723,19 @@ int dinit_main(int argc, char **argv)
...
+                    services->start_service("recovery");
+                }
+                catch (...) {
+                    std::cout << "Unable to start recovery service.\n";
```
davmac314, line 732: This should go to the log, I think.

src/dinit.cc
```diff
@@ -718,7 +723,19 @@ int dinit_main(int argc, char **argv)
...
+                    confirm_restart_boot();
+                }
+            } else {
```
davmac314, line 735: You don't really need the `else` clause, you can just have `confirm_restart_boot()` after the if.

I.e. rather than:

```
if (...) {
    try {
        ...
    }
    catch (...) {
        ...
        confirm_restart_boot();
    }
}
else {
    confirm_restart_boot();
}

```

You can have:

```
if (...) {
    try {
        ...
    }
    catch (...) {
        ...
    }
}
confirm_restart_boot();
```

The extra log message (" Will prompt user.") isn't really needed.

davmac314 (changes requested):

src/dinit.cc
JamiKettunen: The space was to account for https://github.com/davmac314/dinit/blob/927bec0/src/dinit.cc#L686-L697 not handling `shutdown_type_t::NONE` as I saw the log contained the next thing on the same line after `No more active services.` without new line, should that be dealt with or left as-is?

src/dinit.cc
JamiKettunen: The extra log thing was again to end https://github.com/davmac314/dinit/blob/927bec0/src/dinit.cc#L684 with a newline, but perhaps `shutdown_type_t::NONE` could be handled right below or be left as-is anyway, it's just currently missing a linebreak for the next thing appended to log

davmac314: I see, that actually makes sense (and I think in fact you're fixing an existing issue), though it's a bit hard to follow, so I might re-arrange things later. For now then this logic is good, however the `else` should be on the next line to match the code style (see CODE-STYLE).

src/dinit.cc
davmac314: (Ok, this is fine, see my other comment).

davmac314:
One other thing - the man page (doc/manpages/dinit.8.m4) needs to be updated with the new option.

src/dinit.cc
JamiKettunen: I've made it more clear what the intention was by using a more generic `Will handle boot failure.` where the others already were.

davmac314 (approved):
Looks good. Thanks!

davmac314:
I invite contributors to add themselves to the CONTRIBUTORS file if they wish. You are welcome to do this by opening another pull request with that change. Thanks again.

Issue #310, opened 2024-04-02
support for systemd readiness notification mechanism


q66:
We have support for s6-style readiness notification with the pipe/file descriptor. It would be nice to additionally have support for systemd-style readiness notification, at very least for its READY state and perhaps a few others that fit (ignoring the rest). This could be accomplished with something like `ready-notification = socket` or similar.

The mechanism is fairly simple and generic, and would allow seamless support for everything that already implements systemd readiness notification. It's also somewhat more flexible in terms of what can do the notification; since you only need a socket address (carried in through the `NOTIFY_SOCKET` environment variable), it can be easily passed down to children and the likes, without having to worry about keeping track of an open file descriptor.

capezotte:
Would this feature be a Linux exclusive or should we try to implement it for other POSIX OSes?

Since (afaik) only Linux attaches credentials to `SOCK_DGRAM` messages, on other platforms this would imply multiple sockets (possibly with different permissions each if we want to prevent services from impersonating one another).

Besides, the "readiness notification" mechanism has already expanded the scope to fd-holding too, and there's no way to negotiate the supported feature set or to return errors for these calls other than to close the socket since it's one way from the service to the init system.

If we take the "close the socket after getting READY=1" route, we might as well implement something like `dbus-wait-for`/`s6-notifyoncheck` that binds a self-cleaning (abstract) socket, receives and authenticates the message, and exits.

netbsduser:
> Since (afaik) only Linux attaches credentials to `SOCK_DGRAM` messages, on other platforms this would imply multiple sockets (possibly with different permissions each if we want to prevent services from impersonating one another).

Credential passing over datagram sockets is also available on NetBSD, FreeBSD, and DragonFly BSD. OpenBSD does not though some patches, not accepted, seem to have been sent to them to implement it. I am not aware that macOS or Solaris support it, I don't think they do.

PR #401, opened 2024-10-17
Implement systemd-style readiness notification via NOTIFY_SOCKET

q66:
The service may specify this with ready-notification=socket. This will result in dinit creating an abstract datagram socket and perform reads on it. As soon as the READY=1 datagram arrives, the readiness notification is signaled.

Note that this is a very rough proof of concept implementation, e.g. the socket path is hardcoded and so on, I mostly just want to know your opinion on whether to do this at all or how to provide the interface and so on - once I know I should proceed and how, I will clean it up.

Also, as it is it carries a dependency on abstract sockets, which are a Linux extension. I wonder if something else should be done on other platforms, or if to make this an optional Linux-only functionality, or ...

I considered implementing this with a wrapper, but considering the mechanism uses datagram sockets and therefore there is no way to know what the lifetime of the watcher process should be, I decided not to. Doing it in dinit means the service manager knows about how long to keep the socket around.

Fixes https://github.com/davmac314/dinit/issues/310

### DIFF
--- a/src/baseproc-service.cc
+++ b/src/baseproc-service.cc
@@ -30,7 +30,8 @@ void base_process_service::do_smooth_recovery() noexcept
 
 bool base_process_service::bring_up() noexcept
 {
-    if (!open_socket()) {
+    if (!open_socket() || !open_ready_socket()) {
+        becoming_inactive();
         return false;
     }
 
@@ -175,7 +176,7 @@ bool base_process_service::start_ps_process(const std::vector<const char *> &cmd
             }
         }
 
-        if (have_notify) {
+        if (have_notify && ready_socket_fd < 0) {
             // Create a notification pipe:
             if (bp_sys::pipe2(notify_pipe, 0) != 0) {
                 log(loglevel_t::ERROR, get_name(), ": can't create notification pipe: ", strerror(errno));
@@ -185,10 +186,12 @@ bool base_process_service::start_ps_process(const std::vector<const char *> &cmd
             // Set the read side as close-on-exec:
             int fdflags = bp_sys::fcntl(notify_pipe[0], F_GETFD);
             bp_sys::fcntl(notify_pipe[0], F_SETFD, fdflags | FD_CLOEXEC);
+        }
 
+        if (have_notify) {
             // add, but don't yet enable, readiness watcher:
             try {
-                rwatcher->add_watch(event_loop, notify_pipe[0], dasynq::IN_EVENTS, false);
+                rwatcher->add_watch(event_loop, ready_socket_fd >= 0 ? ready_socket_fd : notify_pipe[0], dasynq::IN_EVENTS, false);
                 ready_watcher_registered = true;
             }
             catch (std::exception &exc) {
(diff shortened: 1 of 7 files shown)


q66:
I guess a proper implementation should also do credentials checking and have only *one* socket that is created by dinit on start? Doing this is simpler, but also all sockets in the abstract namespace are available to all processes, which may result in some unfortunate sharing. Doing so with one socket however would require some kind of non-portable checking like via cgroups. I guess using separate sockets and giving them randomized names in the abstract namespace would probably be good enough? hm...

q66:
my main issue with using plain old sockets is the same, i.e. permissions protect them but in the end *any* process running as the same user is allowed to poke its socket to signal readiness, which is kind of unfortunate, but i'm not sure how much of a problem that actually is in practice

q66:
the way systemd gets around it is that it checks the process credentials that come with the datagram, but we can't do that, because the allowed processes should include any child process of the original service process, and there is no way to find out which ones these are - systemd knows because it has a cgroup per service

davmac314:
> my main issue with using plain old sockets is the same, i.e. permissions protect them but in the end any process running as the same user is allowed to poke its socket to signal readiness, which is kind of unfortunate, but i'm not sure how much of a problem that actually is in practice

I would guess it's not a problem in practice at all, but it does make me slightly uneasy too.

q66:
thinking about this more, if i end up doing this at all (which i'm not 100% sure about yet), i will do what you said - i.e. use regular unix domain sockets (and add more fields to configure their permissions and ownership, similar to how the activation socket has them), tell people to use this functionality very sparingly, and require the path to be explicitly specified

it should be enough for the cases where nothing else can be done to get readiness notification, and it shouldn't require anything overly complicated or non-portable...

i don't really like this interface myself either, but i think it's still useful to have (but it's not something we should recommend people to use in general, it's more of a last resort thing)

q66:
I updated the PR to implement things like above. Documentation still pending, but you can review what is there...

capezotte:
> I considered implementing this with a wrapper, but considering the mechanism uses datagram sockets and therefore there is no way to know what the lifetime of the watcher process should be, I decided not to. Doing it in dinit means the service manager knows about how long to keep the socket around.

I'm in favor of using a wrapper. Besides [what I've said in the original issue](https://github.com/davmac314/dinit/issues/310#issuecomment-2169016109), the [`sd_notify` man page](https://www.man7.org/linux/man-pages/man3/sd_notify.3.html#RETURN_VALUE) recommends ignoring errors when using the systemd readiness functions if non-systemd service managers are also targeted. This also allows programs that do require additional communications over the lifetime of the process (such as `RELOADING`, `BARRIER`, fd holding, etc.) to somewhat detect that they're not supported.

IMO, keeping the socket open for the lifetime of the process only helps if we want to try to trick programs that don't follow systemd's own guidelines on how to support a non-systemd use case.

q66:
a wrapper is unreliable regardless, the process may never signal readiness and if an approach like dbus-wait-for is taken you'll end up with a forked off child process that will stay around forever, and having builtin support with a persistent socket adds a fairly minimal amount of code and no dependencies anyway (if that was possible for dbus, i'd prefer to make that builtin too, but it's not), plus it's cleaner and easier to use, and more universal (i've seen services that error when the env is set and the fd is open but the other end is closed)

Issue #398, opened 2024-10-14
capabilities support on linux


q66:
Right now dealing with capabilities is clunky; it would be good to have some degree of built-in support for them. Not yet sure what such support would look like but it's something I want to look into later.

PR #404, opened 2024-10-19
Add capabilities, priority, and oom-score-adjustment support on Linux

q66:
This is not properly tested yet and does not come with documentation. I mostly want to get an opinion on it for now.

This adds 3 things; when dinit is compiled without support, all of them will raise errors:

1) The `capabilities` service field. This is an IAB string; it's parsed with `cap_iab_from_string` so it can take any value that can parse, e.g. `^cap_sys_time,^cap_net_admin` for ambient caps. It supports the `+=` operator, which will append to it, delimiting with a comma.
2) The `secure-bits` service field, which implements securebits flags as a companion to the above.
3) The new option `no-new-privs` which will use `prctl` to prevent gaining new privileges across `execve` (e.g. suid).

For now I want to get an opinion on the overall implementation. One thing I am not sure about for example is the IAB string parsing; I currently do it in `load-service.cc` because it must be done after all the concats have been completed. I find it to be a bit out of place there, however, so maybe there is a better way. Or if you have any other clues about anything, etc...

Fixes https://github.com/davmac314/dinit/issues/398


q66:
I also wonder if i should add functionality for setting scheduling and i/o priorities while at it (both are useful to have especially for services that change uid)? i could also do it later, but it's simple enough to do...

q66:
this is what the setting of the 3 values would look like without the parsing and guards:

```
diff --git a/src/includes/proc-service.h b/src/includes/proc-service.h
index bb7ed42..8d622cd 100644
--- a/src/includes/proc-service.h
+++ b/src/includes/proc-service.h
@@ -43,6 +43,9 @@ struct run_proc_params
     bool on_console;          // whether to run on console
     bool in_foreground;       // if on console: whether to run in foreground
     bool unmask_sigint = false; // if in foreground: whether to unmask SIGINT
+    int nice = INT_MIN;
+    int ionice = INT_MIN;
+    int oom_adj = INT_MIN;
     int wpipefd;              // pipe to which error status will be sent (if error occurs)
     int csfd;                 // control socket fd (or -1); may be moved
     int socket_fd;            // pre-opened socket fd (or -1); may be moved
diff --git a/src/run-child-proc.cc b/src/run-child-proc.cc
index b206b6d..ef9392b 100644
--- a/src/run-child-proc.cc
+++ b/src/run-child-proc.cc
@@ -65,6 +65,9 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
     const char *working_dir = params.working_dir;
     const char *logfile = params.logfile;
     bool on_console = params.on_console;
+    int nice = params.nice;
+    int ionice = params.ionice;
+    int oom_adj = params.oom_adj;
     int wpipefd = params.wpipefd;
     int csfd = params.csfd;
     int notify_fd = params.notify_fd;
@@ -301,6 +304,38 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
         if (setrlimit(limit.resource_id, &setlimits) != 0) goto failure_out;
     }
 
+    // nice
+    if (nice != INT_MIN) {
+        if (setpriority(PRIO_PROCESS, getpid(), nice) != 0) goto failure_out;
+        // we usually create a new session leader; that makes nice not very
+        // useful as the Linux kernel will autogroup processes by session id
+        // except when disabled - so also work around this where enabled
+        // the r+ is used in order to avoid creating it where already disabled
+        errno = 0;
+        FILE *ag = std::fopen("/proc/self/autogroup", "r+");
+        if (ag) {
+            std::fprintf(ag, "%d\n", nice);
+            std::fclose(ag);
+        }
+        else if (errno != ENOENT) goto failure_out;
+    }
+
+    // ionice
+    if (ionice != INT_MIN) {
+        #ifdef __NR_ioprio_set
+        if (syscall(__NR_ioprio_set, 1, (int)getpid(), ionice) != 0) goto failure_out;
+        #endif
+    }
+
+    // oom score adjustment
+    if (oom_adj != INT_MIN) {
+        errno = 0;
+        FILE *adj = std::fopen("/proc/self/oom_score_adj", "w");
+        if (!adj) goto failure_out;
+        std::fprintf(adj, "%d\n", oom_adj);
+        std::fclose(adj);
+    }
+
     #if SUPPORT_CGROUPS
     if (params.run_in_cgroup != nullptr && *params.run_in_cgroup != 0) {
         err.stage = exec_stage::ENTER_CGROUP;
```

q66:
I implemented everything. What's left now is basically

1) documentation
2) testing
3) deciding how to guard the features; whether we should have a separate feature for each mainly
4) the cap parsing thing
5) maybe something else? you tell me

davmac314:
No strong objections/opinions from me, though as usual good documentation will be important.

nice/io_nice/oom_adj are probably unnecessary when cgroups are used properly. If they were to be supported, documentation must clearly state the requirements, i.e. `/proc` must be mounted for `oom_adj`.

> One thing I am not sure about for example is the IAB string parsing; I currently do it in load-service.cc

I think that's fine tbh.

It might be worth gating Linux-specific features behind something other than `__linux__`, so you can use it in documentation too (and so those that prefer to use cgroups could just disable these features).

Setting for nice should  be possible on other OSes, `setpriority` and `nice` are both part of POSIX (and I think you can use `nice` rather than `setpriority` unless I've missed something). The autogroups hack is unfortunate, I guess there's no way around it though.

q66:
well, `nice` will add to the current level, while `setpriority` actually sets it, so there is a difference - both are posix though...

the main non-portability for that is range checking, as the acceptable range of values may vary depending on systems

maybe i could

1) implement `nice` unconditionally - not sure about the range checking part though? `setpriority` will clamp the input, so if you set it too high or too low, it will use the nearest supported value; but either way, do the autogroups hack only for linux
2) put the `ionice` behind its own feature
3) put the `oom-score-adj` behind its own feature

davmac314:
> well, nice will add to the current level, while setpriority actually sets it, so there is a difference

Ah right, yeah.

> the main non-portability for that is range checking, as the acceptable range of values may vary depending on systems

True, but I don't think we need to be overly fussed about checking it ourselves. Perhaps add a flag to indicate whether the value is set rather than relying on INT_MIN as a sentinel value.

> maybe i could [...]

Yep, that sounds reasonable to me.

q66:
I added documentation, feature flags, and improved the resource management for the capabilities IAB object, which should be a lot cleaner now.

It should be reviewable now, though I still need to test it a bunch more.

davmac314:
Will review this weekend

.github/workflows/meson_ci.yml
```diff
@@ -61,12 +61,12 @@ jobs:
       if: ${{ matrix.arch == 'amd64' }}
       run: |
         apt-get update
-        DEBIAN_FRONTEND=noninteractive apt-get install g++ meson m4 -y
+        DEBIAN_FRONTEND=noninteractive apt-get install g++ meson m4 libcap-dev file -y
```
mobin-2008, line 64: Hmm, Why there is a `file`?

src/tests/Makefile
```diff
@@ -3,6 +3,10 @@
 ALL_TEST_CXXFLAGS=$(CPPFLAGS) $(TEST_CXXFLAGS) $(TEST_CXXFLAGS_EXTRA)
 ALL_TEST_LDFLAGS=$(TEST_LDFLAGS) $(TEST_LDFLAGS_EXTRA)
 
+ifeq ($(SUPPORT_CAPABILITIES),1)
+  ALL_TEST_LDFLAGS+=-lcap
+endif
```
davmac314, line 8: This isn't really the right place for this. If tests need `-lcap` then it should be in `TEST_LDFLAGS`. That could be set in `mconfig.Linux`/`mconfig.Linux.sh`, but see comment there.

src/tests/cptests/Makefile
```diff
@@ -3,6 +3,10 @@
 ALL_TEST_CXXFLAGS=$(CPPFLAGS) $(TEST_CXXFLAGS) $(TEST_CXXFLAGS_EXTRA)
 ALL_TEST_LDFLAGS=$(TEST_LDFLAGS) $(TEST_LDFLAGS_EXTRA)
 
+ifeq ($(SUPPORT_CAPABILITIES),1)
+  ALL_TEST_LDFLAGS+=-lcap
+endif
```
davmac314, line 8: (Likewise, this shouldn't be here)

configs/mconfig.Linux
```diff
@@ -38,6 +38,9 @@ TEST_LDFLAGS=$(TEST_LDFLAGS_BASE) $(TEST_CXXFLAGS)
 # Features.
 
 SUPPORT_CGROUPS=1
+SUPPORT_CAPABILITIES=1
```
davmac314, line 41: Given this requires a dependency (in libcap), I'd prefer the default here was `SUPPORT_CAPABILITIES=0`. It's fine to have it on-by-default in the `configure` script since we can check the dependency is available there (but that should be implemented).

src/Makefile
```diff
@@ -9,6 +9,10 @@ ifeq ($(BUILD_SHUTDOWN),yes)
   SHUTDOWN=$(SHUTDOWN_PREFIX)shutdown
 endif
 
+ifeq ($(SUPPORT_CAPABILITIES),1)
+  ALL_LDFLAGS+=-lcap
+endif
```
davmac314, line 14: It should already be `LDFLAGS` if it's needed (see review comments elsewhere for further details).

configure
```diff
@@ -467,6 +482,9 @@ STRIPOPTS=$STRIPOPTS
 # Feature settings
 SUPPORT_CGROUPS=$SUPPORT_CGROUPS
 USE_INITGROUPS=$USE_INITGROUPS
+SUPPORT_CAPABILITIES=$SUPPORT_CAPABILITIES
```
davmac314, line 485: If `SUPPORT_CAPABILITIES` is 1 then `-lcap` needs to be added to `LDFLAGS` and `TEST_LDFLAGS` _if_ they haven't been specified manually. See for example the tests under the heading "`# Check whether sanitizers can be used for tests`" (that's only dealing with `TEST_LDFLAGS` but the general principle should be the same for `LDFLAGS`).

src/includes/proc-service.h
```diff
@@ -14,6 +18,41 @@
...
 class process_service;
 
+#if SUPPORT_CAPABILITIES
+// A thin wrapper around the structure to avoid having to worry about freeing
+// it in various places (move semantics and destructor will take care of it)
+struct cap_iab_wrapper {
+    cap_iab_wrapper() {}
+    cap_iab_wrapper(std::string const &str) noexcept {
+        if (str.empty()) return;
+        iab = cap_iab_from_text(str.c_str());
```
davmac314, line 28: Add a comment that makes it clear that `iab` may still be null after this (if the string is invalid or if memory allocation fails within the `cap_iab_from_text` function).

src/includes/proc-service.h
```diff
@@ -14,6 +18,41 @@
...
+
+    cap_iab_wrapper &operator=(cap_iab_wrapper const &) = delete;
+    cap_iab_wrapper &operator=(cap_iab_wrapper &&v) noexcept {
+        iab = v.iab;
+        v.iab = nullptr;
+        return *this;
+    }
+
+    ~cap_iab_wrapper() noexcept {
+        cap_free(iab);
```
davmac314, line 44: It's not clear from libcap documentation (unless I've missed something) that it is valid to call `cap_free` with a null argument, so I think it is better to guard against that case.

davmac314 (changes requested):
Ok, I actually have a few minor things that I think would be good to change, as noted. Thanks for the PR.

src/includes/load-service.h
```diff
@@ -793,6 +846,30 @@ inline unsigned long long parse_unum_param(file_pos_ref input_pos, const std::st
...
+        const std::string &service_name, long long min = std::numeric_limits<long long>::min(),
+        long long max = std::numeric_limits<long long>::max())
+{
+    const char * num_err_msg = "specified value contains invalid numeric characters or is outside "
+            "allowed range.";
+
+    std::size_t ind = 0;
+    try {
+        long long v = std::stoll(param, &ind, 0);
+        if (v < min || v > max || ind != (param.length() - (v < 0))) {
```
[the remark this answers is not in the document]
q66: yes, this is wrong, i will fix that

src/includes/load-service.h
```diff
@@ -1544,6 +1662,87 @@ void process_service_line(settings_wrapper &settings, const char *name, const ch
...
+                            "secure-bits", input_pos);
+                }
+            }
+            break;
+        }
+        #endif
+        case setting_id_t::NICE:
+        {
+            string nice_str = read_setting_value(input_pos, i, end);
+            settings.nice = 1 | ((int)parse_snum_param(input_pos, nice_str, name,
```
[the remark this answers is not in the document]
q66: the reason i didn't do that is that 1) it actually makes the code more complicated because now you have to check a bunch of booleans around the place, particularly when copying things between the structures and 2) it adds a bunch of pointless structural bloat due to field alignment, and when you take care to put the booleans in one place, it becomes less obvious how they are associated with the rest of the fields

src/tests/Makefile
q66: all linking needs -lcap due to the IAB abstraction in the load-service.h

configs/mconfig.Linux
q66: it may add a dependency, but it's a dependency that's present on virtually *any* linux system (it's required by things like... udev), imo it makes much more sense to keep it on

src/includes/proc-service.h
q66: the implementation explicitly checks it, since 1.95 release in 2007, but sure

meson.build
```diff
@@ -56,6 +59,8 @@ if platform == 'freebsd' and compiler.has_link_argument('-lrt')
     add_project_link_arguments('-lrt', language : 'cpp')
 endif
 
+libcap_dep = dependency('libcap', required: support_capabilities)
```
mobin-2008, line 62: Could you add a comment about this new "Dependencies" Section?
```meson
## Dependencies
libcap_dep = dependency('libcap', required: support_capabilities)
```

meson.build
```diff
@@ -68,6 +73,15 @@ mconfig_data.set10('USE_INITGROUPS', use_initgroups)
 if support_cgroups.auto() and platform == 'linux' or support_cgroups.enabled()
     mconfig_data.set('SUPPORT_CGROUPS', '1') 
 endif
+if libcap_dep.found()
+    mconfig_data.set('SUPPORT_CAPABILITIES', '1')
```
mobin-2008, line 77: This could be:
```meson
mconfig_data.set10('USE_INITGROUPS', use_initgroups)
mconfig_data.set10('SUPPORT_CAPABILITIES', libcap_dep.found());
```

configs/mconfig.Linux
davmac314: The library will be present but note necessarily the headers, and people using a straight Makefile-based build are often just trying out Dinit for personal use. I'd like to make sure that this scenario still works as easily as possible.

src/includes/load-service.h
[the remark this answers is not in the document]
davmac314: There should only be a couple of extra assignments I would think? You don't have to check the booleans, just pass them around along with the nice value. I know what you mean about putting the booleans in one place separate from the other fields they may be related to but it doesn't seem like that much of a deal and is ultimately a lot cleaner.

src/tests/Makefile
davmac314: It can go into both `TEST_LDFLAGS` and `LDFLAGS`, what I mean is that this should happen in the build configuration (mconfig.* / configure script) though rather than being done ad-hoc in various makefiles.

configs/mconfig.Linux
q66: yes, but those people can easily disable it; i don't think it's right to default to a configuration where capabilities are not supported, the default build should imo be one that is as distros should build it

it's trivial to disable for somebody just trying it (it's also trivial to just install libcap-devel or whatever, in 99% of scenarios it's easily doable, and you already need to install a development toolchain)

davmac314: For this build settings template file I'm not so fussed, but at least `-lcap` should be added to `LDFLAGS_BASE` if `SUPPORT_CAPABILITIES=1` is the default. (Or add a separate LIBCAP variable, set it to `-lcap`, and use it in the Makefiles in appropriate places).

For the `configure`-script though, the whole point of the script is to automatically find build settings that will work on the current system as it is.

src/load-service.cc
```diff
@@ -733,6 +741,16 @@ service_record * dirload_service_set::load_reload_service(const char *fullname,
             #if SUPPORT_CGROUPS
             rvalps->set_cgroup(std::move(settings.run_in_cgroup));
             #endif
+            #if SUPPORT_CAPABILITIES
+            rvalps->set_cap(std::move(cap_iab), settings.secbits.get());
+            #endif
+            if (settings.nice_is_set) rvalps->set_nice(settings.nice);
```
davmac314, line 747: To handle service reload properly, this needs to clear the `nice_is_set` value in the service (`rvalps`) if it is clear in the settings that were read.  I suggest making `set_nice` take two parameters: the "is set" flag and the value. The call here then doesn't need to be guarded by `if (settings.nice_is_set)`.

Same for `set_oom_adj`

davmac314 (commented):
It's all good other than the outstanding comments.

configs/mconfig.Linux
mobin-2008: I think I can implement a library finding function based on `pkg-config` for `configure` script.

aanderse:
how do i set `no-new-privs`? simply add to the service file with no `=`? if so... i _think_ this will be the only service property that doesn't use a `=`, right? 

i think it is worthwhile keeping this as a boolean, requiring `no-new-privs = true` because it keeps parsing easier using various libraries... what do you think? 

---

ps. this is great! i am super excited to see capabilities option added ❤️

q66:
it goes in `options =`

aanderse:
oops! ignore me 😅 

thanks again

configs/mconfig.Linux
q66: I misunderstood what you meant the first time. I changed the default mconfigs to disable capabilities, and the configure script detects -lcap for base ldflags and uses it (which is also enough to address the rest)

Since most people are supposed to run configure, I think that's okay

q66:
I think I addressed all the comments, let me know if there's anything else

src/run-child-proc.cc
```diff
@@ -291,6 +311,56 @@ void base_process_service::run_child_proc(run_proc_params params) noexcept
...
+    #if SUPPORT_OOM_ADJ
+    // oom score adjustment
+    if (oom_adj_is_set) {
+        err.stage = exec_stage::SET_PRIO;
+        errno = 0;
+        int fd = open("/proc/self/oom_score_adj", O_WRONLY);
+        if (fd < 0) goto failure_out;
+        // +4: round up, minus sign, newline, nul terminator
+        char val_str[std::numeric_limits<int>::digits10 + 4];
+        int num_chars = snprintf(val_str, sizeof(val_str), "%hd\n", oom_adj);
```
davmac314, line 355: `%hd` should be just `%d`, I believe (the value is a plain `int`, not a `short int`).

davmac314 (changes requested):
Just one tiny format specifier that I think is incorrect, otherwise, looks good. Sorry for taking a little while to get to it.

src/run-child-proc.cc
davmac314: Ah I see. In that case line 354 should have `std::numeric_limits<short>` instead of `std::numeric_limits<int>`.

q66: done

davmac314 (approved):

PR #405, opened 2024-10-20
Fix bad formatting of line number when a service fails to load

JamiKettunen:
It was adding `line X` directly after the service file path without any kind of separation. As an example the following:
```
dinit: Error in service description for 'early-test' (file /etc/dinit.d/early-testline 1): badly formed line.
dinit: Error in service description for 'early-test': setting 'options' (file /etc/dinit.d/early-testline 4): unknown option: invalid
```
would become:
```
dinit: Error in service description for 'early-test' (file /etc/dinit.d/early-test, line 1): badly formed line.
dinit: Error in service description for 'early-test': setting 'options' (file /etc/dinit.d/early-test, line 4): unknown option: invalid
```
after this change.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,6 +12,7 @@ The following people (in alphabetical order) have contributed:
  * Edd Barrett - Testing, documentation.
  * Fabien Poussin - Code, services, documentation.
  * James Knippes - Code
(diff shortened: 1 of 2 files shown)


mobin-2008 (approved):
Thanks, Looks Good.

PR #414, opened 2024-11-24
Fix SVCDSCDIR replies on 64-bit big endian architectures

q66:
While on LE this works, on BE it copies the other four bytes into the uint32_t, resulting in a zero. This in turn results in a protocol error on those architectures.


davmac314 (approved):

davmac314:
Thanks, good catch

PR #447, opened 2025-03-25
meson: Always add trailing slash for IGR_{INPUT/OUTPUT}_BASE paths

mobin-2008:
igr-runner expects this way. meson by default add a trailing slash for current source and build directory but there is no guarantee that it does this way. For example muon doesn't add trailing slash for mentioned paths so always add that slash when it's expected.

### DIFF
--- a/src/igr-tests/meson.build
+++ b/src/igr-tests/meson.build
@@ -1,8 +1,8 @@
 # Included from top-level meson.build
 
 igr_tests_env += [
-    'IGR_OUTPUT_BASE=@0@'.format(meson.current_build_dir()),
-    'IGR_INPUT_BASE=@0@'.format(meson.current_source_dir())
+    'IGR_OUTPUT_BASE=@0@/'.format(meson.current_build_dir()),
+    'IGR_INPUT_BASE=@0@/'.format(meson.current_source_dir())
 ]
 
 runner = executable(


davmac314:
Does IGR_OUTPUT_BASE really need '/' appended? I don't see why

For IGR_INPUT_BASE it looks like the usage was inconsistent in 49284aaf7cc32e8038ad0b7cb4bc7fe6644b5e50.
Eg [this line](https://github.com/davmac314/dinit/commit/49284aaf7cc32e8038ad0b7cb4bc7fe6644b5e50#diff-fbdfee29e321633b2f24339b1ae32fe2c3b4a0f8d2448fa5d3c8c8dc9231a5acR75) appends a '/' (inconsistent with output base),
whereas [this line](https://github.com/davmac314/dinit/commit/49284aaf7cc32e8038ad0b7cb4bc7fe6644b5e50#diff-fbdfee29e321633b2f24339b1ae32fe2c3b4a0f8d2448fa5d3c8c8dc9231a5acR75) expects no '/' at the end,
and [this line](https://github.com/davmac314/dinit/commit/49284aaf7cc32e8038ad0b7cb4bc7fe6644b5e50#diff-51237a8f19e2c09f806916877d149d2306229c63a99fdafb8d225ac8eec0cea5R291) requires it.

I think the problem should be fixed by fixing those inconsistencies in igr-runner. It's unusual that specifying a directory actually requires a slash at the end so I don't think that's the right thing to do. Also I think with the solution you're proposing here it would end up with double-slash at the end if run with meson.

PR #455, opened 2025-04-17
CI: Use ubuntu-24.04 for regular ci

mobin-2008:
Ubuntu-20.04 is getting removed from GitHub CI and we need to move to newer version.

I hope there not be ASLR related bugs.

See https://github.com/actions/runner-images/issues/11101 for more details.

### DIFF
--- a/.github/workflows/regular_ci.yml
+++ b/.github/workflows/regular_ci.yml
@@ -33,7 +33,7 @@ jobs:
 
   Debian-bullseye_build:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     container:
       image: debian:bullseye
     strategy:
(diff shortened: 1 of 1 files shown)


mobin-2008:
We can use `ubuntu-latest` for less maintaining concerns or a specific release to make sure there is no incompatibility between docker images and runner kernel like https://github.com/davmac314/dinit/commit/0c8a10ff71893f3d8efcb5253dcf1677c4e1e961

davmac314:
> We can use `ubuntu-latest` for less maintaining concerns or a specific release to make sure there is no incompatibility between docker images and runner kernel like [0c8a10f](https://github.com/davmac314/dinit/commit/0c8a10ff71893f3d8efcb5253dcf1677c4e1e961)

That commit explicitly switched _from_ `ubuntu-latest` to avoid an issue, and to have a stable CI environment. Switching back to `ubuntu-latest` isn't a good idea.

davmac314 (approved):

PR #458, opened 2025-04-29
meson: Check "linux/ioprio.h" header for ioprio auto case

mobin-2008:
The purpose of `feature` is to provide a auto case for automatic detection so check the availability of `linux/ioprio.h` in auto case.


davmac314 (approved):
This is ok, but I notice the lines in the meson.build file are very long.

In general lines are kept to 110 absolute maximum, it would be good if that could be done here as well.

mobin-2008:
> This is ok, but I notice the lines in the meson.build file are very long.
> 
> In general lines are kept to 110 absolute maximum (100 preferred), it would be good if that could be done here as well.

I will update this PR with lines fixed.

davmac314:
> I will update this PR with lines fixed

This PR has been approved and is ready to merge, please follow normal processes. Open a new PR for new changes.

davmac314:
PR process is documented here: https://github.com/davmac314/dinit/blob/master/doc/PR-PROCESS

PR #459, opened 2025-05-01
Small Modification of Fish Autocomplete Configuration

ZionShao:
Made minor adjustments to the fish auto-completion configuration. Added auto-completion for “enable” and "disable",and now non-root users can view all available services(Previously, non-root users could only complete user services.)


ZionShao:
The commit messages was updated.Please check it

PR #461, opened 2025-05-06
meson: Use static_library for shared sources

mobin-2008:
In Make-based builds tests are build twice with all sources because of different flags for tests. In meson this is not the case so just link to shared source instead of rebuiling over and over. Also this fixes building every test source for every single test target.

This reduces ninja targets count from 98 to just 60.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,26 +1,39 @@
 # Included from top-level meson.build
 
 ## Includes
-# We include `../`, `./includes/` & `../dasynq/include/` as defualt_incdir
 default_incdir = include_directories(
     '../',
     'includes/',
     '../dasynq/include/'
 )
-dinit_source_files = [
-    'dinit-main.cc',
-    'dinit.cc',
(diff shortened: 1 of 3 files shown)


davmac314:
> In Make-based builds tests are build twice with all sources because of different flags for tests. In meson this is not the case

The tests are supposed to be built with different flags. Why isn't the meson build doing that? It should be fixed if not.

davmac314:
Another meson issue, each integration test _still_ needs to be listed out in the meson.build file - that's not sustainable. [As I said some time ago](https://github.com/davmac314/dinit/pull/272#pullrequestreview-1807909204):

> As I've suggested before, the right way to run the integration tests is via igr-runner itself. If you really want to run the tests individually from the meson build then the best solution is probably to make changes to igr-runner necessary to allow it to list available tests and to run an individual test, and then have the meson build use that functionality.

That needs to be sorted out soon. It's been long enough.

PR #495, opened 2025-08-31
shutdown: ignore 'now' argument

### LINKED ISSUE #494: Add support for poweroff time, or at least not error out on "now"
**We do not usually accept feature requests, please see the README. You may open a feature request
if you are a distribution/package maintainer and have a clear need for the feature, or if you are
willing to develop the feature yourself and would like to discuss it first.**
It shouldn't be hard, I can hack around with C if you want.

**Your background**
A developer of a script that uses kexec.

**The problem**
`kexec` depends on the command: `shutdown -r now` to reboot the system, NOW!
But dinit's shutdown command does not support `now` as an argument so it errors out instead of rebooting.
Actually dinit's shutdown command does not support any timing at all, it just does it NOW!

**The solution**
Add in the shutdown command source code: if you found a stray "now" argument, just ignore it.

**Additional context**
See: https://github.com/horms/kexec-tools/blob/daa29443819d3045338792b5ba950ed90e79d7a5/kexec/kexec.c#L937

xqrs:
Fixes #494

### DIFF
--- a/src/shutdown.cc
+++ b/src/shutdown.cc
@@ -320,6 +320,10 @@ int main(int argc, char **argv)
             }
         }
         else {
+            // Ignore a single "now" argument. I.e., "shutdown -r now".
+            if (i == argc-1 && strcmp(argv[i], "now") == 0)
+                break;
+
             // time argument? TODO
             show_help = true;
         }


davmac314:
As per the PR template:

```
You *must* read the CONTRIBUTING guide (doc/CONTRIBUTING) before opening a Pull Request.

Do not proceed before reading that guide and referenced files.

```

I don't believe you've done that - the canary text is missing, the code style is incorrect, the documentation isn't updated.

You would need to follow the style guidelines and adjust documentation appropriately, at least.

Finally, did you test this change? I don't believe it will actually work. "shutdown -r now" is just going to do a reboot, not execute the loaded-for-kexec kernel.

PR #504, opened 2025-10-31
configure: fix incomplete message for -fno-rtti check against compiler

mobin-2008:
Fix missing quote for "Checking whether \"-fno-rtti\" is accepted by the compiler..." message.

Since https://github.com/davmac314/dinit/commit/36e3ecf98c4ce5196de3bddf6727ca0f1eb62332 all messages should be quoted, but the commit missed this message, resulting in only "Checking" being printed for this check.


davmac314 (approved):
Weird that I didn't see this

Issue #508, opened 2025-11-05
Print service description file path as part of "dinitctl status" output


iacore:
## Background

I contributed to dinit in the past. I want to discuss about this first.

## Request

I want either `dinitctl status` or `dinitctl info --path` (new command!) to print the service description file path.

Neither `dinitcheck` nor `dinitctl` will print the service file path it is loaded from. For a system admin, this feature is useful.

davmac314:
I'd be happy enough for `dinitctl status` to print the path. The underlying functionality is there, this would only require changes in `dinitctl` and not in `dinit` itself.

As per the contributing guidelines code changes must be done in consistent style and quality as existing code. I don't have time to mentor if your skills aren't up to the task, but if they are, this would be a suitable change to be included before the 1.0 release.

iacore:
I made a simple program in the past.

https://git.sr.ht/~iacore/dinit-edit/tree/main/item/main.zig

`dinit-edit` lists all service files.

```
❯ dinit-edit 
boot              /home/user/.config/dinit.d/boot
...
```

`dinit-edit <SERVICE>` opens the service file in default editor.

Should `dinitctl list` show the file path as well?

Maybe `dinit-edit` could be included as a separate program with dinit.

davmac314:
> Should dinitctl list show the file path as well?

No, at least not by default.

> Maybe dinit-edit could be included as a separate program

You can open a PR to include it under the `contrib` folder if you like.

davmac314:
> [@davmac314](https://github.com/davmac314) I don't think the current use of `cout <<` represents good quality code... is there a better way to output strings?

As I mentioned above, contributed code must be consistent with existing code - which outputs via `cout <<`.

> I don't know how to get the service filename from dinit (daemon). If the list of directories dinit refers to are consistent, 

There is a request to get the service description directory for a service (by handle) and even a function in dinitctl already implemented (`get_service_description_dir`). You can append the name of the service to the directory to get the file.

> wouldn it be possible for dinitctl to find the service files on its own?

That runs risks of being affected by changing system state. It would be better to show where the service was actually loaded from, not where it would be loaded from if it were to be reloaded now.

PR #510, opened 2025-11-09
Add dinit-edit, a simple tool that lists and edits service files

iacore:
Idea from https://github.com/davmac314/dinit/issues/508#issuecomment-3506476353

Solves:

- "Where are my service files?"
- Quickly edit a existing service file

The set of directories searched comes from dinit's man page.

(see patch) Is this the proper way to add a license file for a subdirectory?

### DIFF
--- a/contrib/dinit-edit/COPYING
+++ b/contrib/dinit-edit/COPYING
@@ -1,0 +1,13 @@
+        DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE 
+                    Version 2, December 2004 
+
+ Copyright (C) 2004 Sam Hocevar <sam@hocevar.net> 
+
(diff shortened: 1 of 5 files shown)


davmac314:
> (see patch) Is this the proper way to add a license file for a subdirectory?

That is fine.

Could you add a note in the readme explaining that there are two implementations (Go and Zig) before giving instructions on how to build them both?

Please add "contrib:" at the beginning of the commit message to make clear this is an extra, not part of core Dinit.

davmac314:
> Well, you still can change the commit message now.

Technically, I could change the comment now, but I have merged the changes already and changing it now would be rewriting the branch history, which is (rightfully) considered bad practice.

PR #518, opened 2025-12-09
CI: Update FreeBSD 14.2 to FreeBSD 14.3

mobin-2008:
The support for FreeBSD 14.2-RELEASE ended on September 30, 2025, and Cirrus CI has removed its 14.2 instance.

See https://www.freebsd.org/security/unsupported/ for unsupported releases information.

This fixes the recent failure of FreeBSD CI in commit 0e6a14f.


davmac314 (approved):
Remember that commit comments are supposed to be written in the imperative.

> Write in the imperative, i.e. the message should be able to follow "This change will ..."

PR #527, opened 2026-01-13
Add dinitctl reexec command to re-execute dinit binary

### LINKED ISSUE #526: Feature request: dinitctl reexec - re-execute dinit binary
## Use Case

When running a system with a read-only `/usr` mounted from a squashfs image, it's desirable to hot-reload `/usr` with a new squashfs without a full reboot. The workflow is:

1. Stop all services (`dinitctl stop boot`)
2. Mount new squashfs, swap `/usr` mount (lazy unmount old)
3. Re-execute dinit from the new `/usr/bin/dinit`
4. Start services (`dinitctl start boot`)

Currently step 3 is not possible. The old dinit binary (PID 1) continues running from memory, holding the old squashfs open (preventing full unmount). While lazy unmount works around this, the old squashfs remains in memory until reboot.

Other init systems support this:
- systemd: `systemctl daemon-reexec`
- s6: `s6-svscanctl -a`

## Proposed Solution

Add a `reexec` subcommand to `dinitctl` that instructs dinit to re-execute itself.

### Minimal Implementation

**dinit side:**
1. Add new control protocol command (e.g., `DINIT_CP_REEXEC`)
2. Handler that calls `execve(argv[0], argv, environ)`

**dinitctl side:**
1. Add `reexec` subcommand that sends the command

### Behavior

- `dinitctl reexec` - dinit re-execs using its original `argv[0]` and full original `argv`
- No state serialization needed for the minimal case (user stops services before reexec)

### Why argv[0]?

Using the original `argv[0]` as the binary path is the simplest and most consistent approach. If we're preserving `argv[1..n]` (the flags like `--system`), there's no reason not to also use `argv[0]` as the default path. This means if dinit was started as `/usr/bin/dinit --system`, it re-execs exactly that.

The only edge case would be relative paths or PATH lookup, but for PID 1 that's essentially never the case - the kernel/bootloader always uses an absolute path.

## Future Enhancements (not required for initial implementation)

- Optional path argument: `dinitctl reexec [path]` to exec a different binary
- State serialization for seamless service continuity
- Preserving service states across reexec

## Summary

The minimal implementation is straightforward:
1. Receive reexec command on control socket
2. `execve(argv[0], argv, environ)`

This would enable hot-reload workflows for immutable/image-based systems using dinit.

hholst80:
## Summary

This adds a new `dinitctl reexec` command that instructs dinit to re-execute itself using its original `argv[0]` and arguments.

- Adds `REEXEC` shutdown type and control protocol command
- Implements `dinitctl reexec` subcommand
- On reexec failure, falls back to reboot

## Use case

When running a system with a read-only `/usr` mounted from a squashfs image, this enables hot-reloading `/usr` with a new squashfs without a full reboot:

1. Stop all services (`dinitctl stop boot`)
2. Mount new squashfs, swap `/usr` mount (lazy unmount old)
3. Re-execute dinit from the new `/usr/bin/dinit` (`dinitctl reexec`)
4. Start services (`dinitctl start boot`)

This is a minimal implementation that does not include state serialization - users should stop services before re-exec.

Similar functionality exists in other init systems:
- systemd: `systemctl daemon-reexec`
- s6: `s6-svscanctl -a`

Closes #526

## Test plan

- [ ] Build and verify compilation succeeds
- [ ] Test `dinitctl reexec` triggers re-execution
- [ ] Verify fallback to reboot on exec failure

### DIFF
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -66,6 +66,7 @@ static void control_socket_ready() noexcept;
 static void confirm_restart_boot() noexcept;
 static void flush_log() noexcept;
 static void do_soft_reboot(char **argv) noexcept;
+static void do_reexec(char **argv) noexcept;
 
 static void control_socket_cb(eventloop_t *loop, int fd) noexcept;
 
@@ -707,6 +708,9 @@ int dinit_main(int argc, char **argv)
         else if (shutdown_type == shutdown_type_t::KEXEC) {
             log_msg_end(" Will kexec.");
         }
+        else if (shutdown_type == shutdown_type_t::REEXEC) {
+            log_msg_end(" Will re-exec.");
+        }
         else if (shutdown_type == shutdown_type_t::NONE) {
             log_msg_end(" Will handle boot failure.");
         }
@@ -722,6 +726,15 @@ int dinit_main(int argc, char **argv)
 
             // if we get here, soft reboot failed; reboot normally
             log(loglevel_t::ERROR, "Could not soft-reboot. Will attempt reboot.");
+            shutdown_type = shutdown_type_t::REBOOT;
+            need_log_flush = true;
+        }
+
+        if (shutdown_type == shutdown_type_t::REEXEC) {
+            do_reexec(argv);
+
+            // if we get here, reexec failed
+            log(loglevel_t::ERROR, "Could not re-exec. Will attempt reboot.");
             shutdown_type = shutdown_type_t::REBOOT;
             need_log_flush = true;
         }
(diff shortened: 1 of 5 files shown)


src/dinit.cc
```diff
@@ -726,6 +730,15 @@ int dinit_main(int argc, char **argv)
             need_log_flush = true;
         }
 
+        if (shutdown_type == shutdown_type_t::REEXEC) {
+            do_reexec(argv);
+
+            // if we get here, reexec failed
+            log(loglevel_t::ERROR, "Could not re-exec. Will attempt reboot.");
+            shutdown_type = shutdown_type_t::REBOOT;
```
Wateir, line 738: Why reboot if fail ? if the user choose reexec rather than reboot, it's because they don't want a reboot, for x or y reason ? So why force a reboot ? better to just inform the user of the fail, and suggest it to reboot then

davmac314: Note that Wateir is not a maintainer, this has not been reviewed by a maintainer yet.

@Wateir it's fine to comment on a PR but please make it clear that you are not a maintainer doing a formal review.

hholst80: I agree on the opinionated stance of the fallback reboot. dinit has a nice prompt we could just fall back to giving the regular options here in an exceptional case.

davmac314:
Closing, as per discussion on #526.

PR #545, opened 2026-05-08
dinit-check: fix null dereference on "consumer-of: non-existent" and an igr-test for it

mobin-2008:
I was writing a service for my Dinit instance and then I saw a segfault in the output of the dinit-check command. For example:
```
Checking service: boot...
Checking service: non-existent...
Unable to load service 'non-existent': service description not found.
Performing secondary checks...
Warning: Service 'boot' specified as consumer of service 'Segmentation fault         dinit-check -d src/igr-tests/check-missing-consumer/sd
```
I fixed the null dereference issue causing the segfault, see the first commit and also included a igr-test to make sure it always works correctly.


davmac314:
Could you add some more explanation in the first commit comment please - eg explain at least that the null pointer dereference occurs while printing the warning

src/igr-tests/igr-runner.cc
```diff
@@ -52,15 +53,17 @@ int main(int argc, char **argv)
...
+            { "check-consumer-of-non-existent", check_consumer_of_non_existent_test },
+            { "reload1", reload1_test }, { "reload2", reload2_test },
+            { "no-command-error", no_command_error_test }, { "add-rm-dep", add_rm_dep_test },
+            { "var-subst", var_subst_test }, { "svc-start-fail", svc_start_fail_test, },
+            { "dep-not-found", dep_not_found_test }, { "pseudo-cycle", pseudo_cycle_test },
+            { "before-after", before_after_test }, { "before-after2", before_after2_test },
+            { "log-via-pipe", log_via_pipe_test }, { "catlog", catlog_test },
+            { "offline-enable", offline_enable_test }, { "xdg-config", xdg_config_test },
+            { "cycles", cycles_test }, { "svc-arg", svc_arg_test },
+            { "svc-arg-consumer", svc_arg_consumer_test },
```
davmac314, line 66: Rather than change all these lines, just add the new test at the end.

davmac314 (changes requested):

davmac314:
> Could you add some more explanation in the first commit comment please - eg explain at least that the null pointer dereference occurs while printing the warning

Also in the summary, take "non-existing" out of the quotes, since it's not literal. Eg:

`dinit-check: fix segfault on non-existing "consumer-of" service`

src/igr-tests/igr-runner.cc
mobin-2008: Do I need to put the new test at the end in other places like function declaration and actual test function? or just for this list?

davmac314: If the list here matches the order in the source code / declarations then: yes, if you move it here, you should move it there as well.

However, looking at this again I guess you've put it where you have because it keeps the dinit-check -based tests together? In that case, this is ok to leave as-is. But the commit comment could use improvement as I mentioned.

mobin-2008: > If the list here matches the order in the source code / declarations then: yes, if you move it here, you should move it there as well.
>
> However, looking at this again I guess you've put it where you have because it keeps the dinit-check -based tests together? In that case, this is ok to leave as-is. But the commit comment could use improvement as I mentioned.

Yes, Exactly.

You mean for the first commit? yes, I improved it and I will push the changes after all things get resolved (i.e. when there is nothing to talk about and next round of review process can begin).

davmac314: It is ok, only the commit comment is my concern.

mobin-2008: So I'm going to push the changes.

davmac314:
Ok, there are a couple of minor grammar errors:
> Get the value of "consumer-of" property by using the service name
instead of using the service pointer in this context.

value of _the_ "consumer-of" property

> When dinit-check tries to print the error message for using
non-existing "consumer-of" service, it uses the service pointer which
is null.

for using _a_ non-existing "consumer-of" service

But also, the first paragraph says "in this context" but the context isn't given until the 2nd paragraph, you should re-arrange/reword so that it is clear, thanks.

mobin-2008:
> Ok, there are a couple of minor grammar errors:
> 
> > Get the value of "consumer-of" property by using the service name
> > instead of using the service pointer in this context.
> 
> value of _the_ "consumer-of" property
> 
> > When dinit-check tries to print the error message for using
> > non-existing "consumer-of" service, it uses the service pointer which
> > is null.
> 
> for using _a_ non-existing "consumer-of" service

Fixed.

> But also, the first paragraph says "in this context" but the context isn't given until the 2nd paragraph, you should re-arrange/reword so that it is clear, thanks.

I merged those two paragraphs into one to provide a ~more~ clearer explanation.

davmac314:
Use punctuation to break it up into pieces that aren't too large, and re-order so that it is easy to understand. Eg.

"When printing the error message for a non-existing "consumer-of" service, print the non-existing service name directly using the consumer_of_name field rather than trying to retrieve it via the service pointer (which will be null in this case)."

davmac314:
(you can use that exactly if you want, or rewrite it yourself. Make sure you read your messages back to yourself, it should be pretty clear that what you have now is too long without pauses).

davmac314:
Here's a version that's closer to what you had, which would also be fine:

"Get the value of the "consumer-of" property by using the service name field instead of using the service pointer, when printing the error message about a non-existing service specified in the service description. The pointer is null in this case since the service doesn't exist."

davmac314 (approved):

PR #568, opened 2026-07-22
dinit: swap incorrectly named variables

alliumm:
Previously `am_system_mgr` was set when dinit is the init process and `am_system_init` was set when dinit is run as root. This is confusing so I swapped them.


davmac314:
The names might be confusing in some ways but they are correct. `am_system_mgr` is set if dinit acts as system manager (manages some system startup functions and shutdown/reboot), `am_system_init` is set if dinit uses the sytem paths for service description and configuration files.

You must read [doc/CONTRIBUTING](https://github.com/davmac314/dinit/blob/master/doc/CONTRIBUTING) before opening a pull request. This is mentioned in the pull request template. Please read the appropriate project documentation if you would like to contribute in future.

Issue #558, opened 2026-06-26
Run CI on FreeBSD again


mobin-2008:
I wanted to work on configure fixes after the #550 but I realized that we lost our FreeBSD testing and I want to bring it back first.

Previously we used Cirrus CI for our FreeBSD testing, however [Cirrus Labs joined OpenAI and shut down Cirrus CI in June 1st, 2026](https://cirruslabs.org/).

I had a little experiment with GitHub actions and now I have a working example of FreeBSD testing inside GitHub that

1. Uses KVM acceleration instead of QEMU `TCG` JIT compiler which yields near-native performance and I hope more accurate results
2. Only takes ~4 minutes from start to finish
3. Doesn't depend on any third-party actions except for the official `actions/checkout` and `actions/upload-artifacts`

Take a look at the example run of it: https://github.com/mobin-2008/dinit/actions/runs/28232281153/job/83638673157

davmac314:
It would be good to have FreeBSD in CI, but I think we may as well make use of the existing `freebsd-vm` action (https://github.com/marketplace/actions/freebsd-vm) which will handle caching of the VM image and in general simplify the CI script a lot.

Edit: there's also`cross-platform-actions` (https://github.com/cross-platform-actions/action) which can do other operating systems as well (and also caches images). It might be nice to have an OpenBSD or NetBSD build as well. (We can't have too many or we'll hit the cache limit).

PR #571, opened 2026-08-01
CI: Add FreeBSD, OpenBSD and NetBSD builds

mobin-2008:
Introduce CI builds for three BSD operating systems using cross-platform-actions, based on QEMU virtual machines accelerated with KVM for near-native performance.

The FreeBSD build enables additional Clang hardening checks to detect issues such as integer overflow and to improve memory safety.

Changes are self-reviewed and tested in my fork: https://github.com/mobin-2008/dinit/actions/runs/30704553393/job/91381298660

Addresses #558

### DIFF
--- a/.github/workflows/regular_ci.yml
+++ b/.github/workflows/regular_ci.yml
@@ -144,3 +144,79 @@ jobs:
       with:
           name: igr-tests_output
           path: src/igr-tests/*/output/*
+
+  BSD_build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        # By default all "steps" commands are run by "bash" on the runner itself. BSD builds run
+        # inside virtual machines and cpa.sh is a helper to run commands of "steps" inside the
+        # virtual machine, replacing the default "bash".
+        shell: cpa.sh {0}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: 'FreeBSD'
+            version: '15.1'
+            arch: 'amd64'
+            hardening: 'clang_hardening'
+          - name: 'OpenBSD'
+            version: '7.9'
+            arch: 'amd64'
+          - name: 'NetBSD'
+            version: '10.1'
+            arch: 'amd64'
+
+    steps:
+    - uses: actions/checkout@v7
+    - name: Run VM
+      uses: cross-platform-actions/action@v1.3.0
+      with:
+        operating_system: ${{ matrix.name }}
+        version: ${{ matrix.version }}
+    - name: Install required packages (FreeBSD)
+      if: ${{ matrix.name == 'FreeBSD' }}
+      run: |
+        sudo pkg update
+        sudo pkg install -y gmake m4 llvm
+    - name: Install required packages (OpenBSD)
+      if: ${{ matrix.name == 'OpenBSD' }}
+      run: |
+        sudo pkg_add -I gmake m4 llvm%21
+    - name: Install required packages (NetBSD)
+      if: ${{ matrix.name == 'NetBSD' }}
+      run: |
+        sudo pkgin update
+        sudo pkgin -y install gmake m4 clang
+    - name: Print clang++ architecture and version
+      run: clang++ --version
+    - name: Configure with Clang hardening flags
+      if: ${{ matrix.hardening == 'clang_hardening' }}
+      run: |
+        ./configure CXXFLAGS="-std=c++11 -flto -fstack-clash-protection -fvisibility=hidden \
+        -fsanitize=cfi,address,signed-integer-overflow,integer-divide-by-zero \
+        -fsanitize-trap=cfi,signed-integer-overflow,integer-divide-by-zero \
+        -fno-sanitize-recover" \
+        LDFLAGS="-flto -fsanitize=cfi,address,signed-integer-overflow,integer-divide-by-zero \
+        -fsanitize-trap=cfi,signed-integer-overflow,integer-divide-by-zero"
+    - name: Build
+      run: gmake CXX=clang++
+    - name: Unit tests
+      run: gmake check
+    - name: Integration tests
+      id: igr-tests
+      run: gmake check-igr
+    # cpa.sh doesn't sync files when the step fails. We usually don't need to sync files after a
+    # command failure but we need to sync for the igr-tests output on failure. Explicitly sync
+    # files from the VM to the runner for this case.
+    - name: Sync igr-tests output file(s) from the VM on failure
+      if: ${{ failure() && steps.igr-tests.outcome == 'failure' }}
+      shell: /bin/sh {0}
+      run: cpa.sh --sync-files vm-to-runner
+    - name: Upload igr-tests output file(s) on failure
+      uses: actions/upload-artifact@v6.0.0
+      if: ${{ failure() && steps.igr-tests.outcome == 'failure' }}
+      with:
+        name: igr-tests_output
+        path: src/igr-tests/igr-output/


.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
       with:
           name: igr-tests_output
           path: src/igr-tests/*/output/*
+
+  BSD_build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: cpa.sh {0} --environment-variables CXX
```
davmac314: I had to look up documentation to find out what "cpa.sh" is and does. It should have comments to explain (at least briefly) what it is and where to find more information.

Is `--environment-variables CXX` necessary? Wouldn't it be cleaner to specify the `environment_variables` input to the action?

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+    env:
+      # Make "configure" always use clang++. Our NetBSD image has g++ 10 pre-installed, and
+      # "configure" prefers it over clang++, but g++ 10 is too old to build Dinit.
+      # Note that it's defined here for all the BSDs because it's cleaner than specifying it only
+      # for NetBSD. FreeBSD already uses clang++ due to the lack of g++ in its image and OpenBSD
+      # uses a static mconfig that specifies clang++ as the C++ compiler.
+      CXX: clang++
```
davmac314: Specifying CXX as clang++ via the build (and/or configure) command line would be simpler than forcing it into the environment, wouldn't it?

In "Note that it's defined here" why is this a "Note" and what is "it"?

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+    strategy:
+      fail-fast: false # Be able to upload src/igr-tests/igr-output/ files in igr-tests step
```
davmac314: I don't think the comment accurately describes what this is needed for.

See documentation here: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+            arch: 'amd64'
+
+    steps:
+    - uses: actions/checkout@v7
+    - name: Run VM
+      uses: cross-platform-actions/action@v1.3.0
+      with:
+        operating_system: ${{ matrix.name }}
+        version: ${{ matrix.version }}
+    - name: Getting dependencies (FreeBSD)
```
davmac314: Rather than "Getting dependencies" maybe "Installing required packages". It is more descriptive and more accurate. The packages being installed are not dependencies of Dinit.

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+        ./configure CXXFLAGS="-std=c++11 -flto -fstack-clash-protection -fvisibility=hidden \
+        -fsanitize=cfi,address,signed-integer-overflow,integer-divide-by-zero \
+        -fsanitize-trap=cfi,signed-integer-overflow,integer-divide-by-zero \
+        -fno-sanitize-recover" \
+        LDFLAGS="-flto -fsanitize=cfi,address,signed-integer-overflow,integer-divide-by-zero \
+        -fsanitize-trap=cfi,signed-integer-overflow,integer-divide-by-zero"
+    - name: Build
+      run: gmake
+    - name: Print dinit executable file architecture
+      run: file ./src/dinit
```
davmac314: Isn't this redundant considering `clang++ -dumpmachine`/`clang++ --version` above?

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+        sudo pkg_add -I gmake m4 llvm%21
+    - name: Getting dependencies (NetBSD)
+      if: ${{ matrix.name == 'NetBSD' }}
+      run: |
+        sudo pkgin update
+        sudo pkgin -y install gmake m4 file clang
+    - name: Print clang++ architecture and version
+      run: |
+        clang++ -dumpmachine
+        clang++ --version
```
davmac314: Does `clang++ -dumpmachine` display anything that `clang++ --version` doesn't already?

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+      run: gmake
+    - name: Print dinit executable file architecture
+      run: file ./src/dinit
+    - name: Unit tests
+      run: gmake check
+    - name: Integration tests
+      run: gmake check-igr
+    - name: Upload igr-tests output file(s) on failure
+      uses: actions/upload-artifact@v6.0.0
+      if: failure()
```
davmac314: Doesn't this mean igr-tests output is uploaded if any prior step fails even if the integration tests pass?

davmac314 (changes requested):

.github/workflows/regular_ci.yml
mobin-2008: No. `-dumpmachine` outputs the target machine and the `--version` outputs that too with some other information as well. `clang++ -dumpmachine` Will be removed.

mobin-2008: (but g++ only outputs version and copyright information on `--version` flag)

.github/workflows/regular_ci.yml
mobin-2008: The original use case for this was to make sure that we are building 32-bit binary on Debian i386 target. Considering that g++ `-dumpmachine` or clang++ `--version` shows the architecture, Yeah, this can be removed and I will remove it.

.github/workflows/regular_ci.yml
mobin-2008: Added a comment to explain what is `cpa.sh` and why it's needed.

`environment_variables` seems cleaner. Will be fixed.

.github/workflows/regular_ci.yml
mobin-2008: The only build that has a 'Configure' step in it is the FreeBSD build and that's for specifying compiler and linker flags. NetBSD also uses configure by default due to lack of static mconfig for NetBSD. For explicitly passing compiler to `configure`, we need to specify an extra step with a condition of `${{ matrix.name == 'NetBSD' }}` which I think is not cleaner than specifying this environment variable here.

If we go with passing compiler to Make directly, every step invoking Make should change to include `CXX=clang++` which also I think is not cleaner than this approach.

> In "Note that it's defined here" why is this a "Note" and what is "it"?

That's because I use to have a "Note" for pointing out a thing (especially in my native language writings) but it's not needed here and only hurts the flow of the paragraph. I will remove the 'Note that'.
'it' is the environment variable here. Will change to explicitly say 'This environment variable is...'.

.github/workflows/regular_ci.yml
mobin-2008: That is a odd case to have any prior step fail but integration tests pass. If build fails, that's not going to happen. If unit tests fail, that's unlikely to happen.

I can modify this condition to check for igr-tests [outcome](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context) to be `failure` before trying to upload.

.github/workflows/regular_ci.yml
davmac314: > For explicitly passing compiler to configure, we need to specify an extra step

No, `make CXX=clang++` runs configure with CXX set to `clang++`, doesn't it? Once that's done you don't need to pass `CXX=clang++` to further `make` invocations.

But, you can keep it in the environment if you prefer.

> 'it' is the environment variable here. Will change to explicitly say 'This environment variable is...'

Yes, you'll need to fix that. You can only use "it" to refer to something that is already being discussed.

.github/workflows/regular_ci.yml
davmac314: Ok.

.github/workflows/regular_ci.yml
davmac314: > That is a odd case to have any prior step fail but integration tests pass.

Well actually, as it is currently no steps will continue if a prior one fails - except this one. So if the build or unit tests fail for example, integration tests won't even run, but this step would still be run.

> I can modify this condition

Yes.

.github/workflows/regular_ci.yml
davmac314: > clang++ -dumpmachine Will be removed

Ok.

.github/workflows/regular_ci.yml
davmac314: Ok.

.github/workflows/regular_ci.yml
mobin-2008: In that case I think removing it is a good idea.

davmac314: Do you mean removing the comment, or the setting itself?

mobin-2008: Removing the comment.

I think the setting is still useful, we don't want other builds based on the same matrix fail because of a single build failure.

davmac314: Ok.

mobin-2008:
I think there are no questions left and I'm pushing the changes to be reviewed.

In addition to normal workflow tests in the PR, I tested the igr-tests failure path in another branch to make sure it works properly. It's tested here: https://github.com/mobin-2008/dinit/actions/runs/31600231519/job/94125510276

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+    - name: Integration tests
+      id: igr-tests
+      run: gmake check-igr
+    # cpa.sh doesn't sync files when the step fails. We usually don't need to sync files after a
+    # command failure but we need to sync for the igr-tests output on failure. Explicitly sync
+    # files from the VM to the runner for this case.
+    - name: Sync igr-tests output file(s) from the VM on failure
+      if: ${{ failure() && steps.igr-tests.outcome == 'failure' }}
+      shell: /bin/sh {0}
+      run: cpa.sh --sync-files vm-to-runner
```
davmac314, line 216: If you make changes outside of what was discussed in review, you should make this clear in a discussion comment on the PR. You need to communicate properly with the reviewer.

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+    - name: Unit tests
+      run: gmake check
+    - name: Integration tests
+      id: igr-tests
+      run: gmake check-igr
+    # cpa.sh doesn't sync files when the step fails. We usually don't need to sync files after a
+    # command failure but we need to sync for the igr-tests output on failure. Explicitly sync
+    # files from the VM to the runner for this case.
+    - name: Sync igr-tests output file(s) from the VM on failure
+      if: ${{ failure() && steps.igr-tests.outcome == 'failure' }}
```
davmac314: In `failure() && steps.igr-tests.outcome == 'failure'` isn't `failure() &&` redundant?

.github/workflows/regular_ci.yml
```diff
@@ -144,3 +144,79 @@ jobs:
...
+      shell: /bin/sh {0}
+      run: cpa.sh --sync-files vm-to-runner
+    - name: Upload igr-tests output file(s) on failure
+      uses: actions/upload-artifact@v6.0.0
+      if: ${{ failure() && steps.igr-tests.outcome == 'failure' }}
```
davmac314: Another redundant `failure() &&`.

davmac314 (approved):

.github/workflows/regular_ci.yml
mobin-2008: Yes. I will comment on any change outside of what has been discussed.

Issue #552, opened 2026-06-06
'configure' does not write XXX_FOR_BUILD flags to mconfig if variables are set but empty


davmac314:
The configure script should always write the (possibly empty) values of CXXFLAGS_FOR_BUILD, CPPFLAGS_FOR_BUILD and LDFLAGS_FOR_BUILD to the mconfig file if they are set.

Currently it doesn't write the values if they are empty. The makefiles then fall back to using the regular (not _FOR_BUILD) flags, resulting in incorrect flags being used.

PR #578, opened 2026-08-19
configure: write empty-but-set XXX_FOR_BUILD flags to mconfig

mobin-2008:
Always write 'CXXFLAGS_FOR_BUILD', 'CPPFLAGS_FOR_BUILD' and 'LDFLAGS_FOR_BUILD' variables to mconfig if they are set.

Addresses #552

Tested with `./configure CXX="clang++" CXXFLAGS="--target=aarch64-linux-gnu -std=c++11" CXXFLAGS_FOR_BUILD=""`

### DIFF
--- a/configure
+++ b/configure
@@ -822,13 +822,13 @@ if [ -n "${CXX_FOR_BUILD:-}" ]; then
         echo "CXX_FOR_BUILD=$CXX_FOR_BUILD"
     } >> mconfig
 fi
-if [ -n "${CXXFLAGS_FOR_BUILD:-}" ]; then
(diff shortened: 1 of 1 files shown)


davmac314 (changes requested):
The change is fine, but can you please adjust the commit message slightly so that it reads clearly: hyphenate 'empty-but-set'.

    configure: write empty-but-set XXX_FOR_BUILD flags to mconfig

Otherwise it reads like "write empty, but set XXX_FOR_BUILD flags to mconfig" which doesn't make sense.

davmac314 (approved):

PR #582, opened 2026-08-31
control: add missing return after BADREQ in add/rm dep

nikitax44:
Firstly, I have read and understand the CONTRIBUTING guide in full.

I'm currently building systemd-like nix API on top of dinit.
While implementing a forward-port of `SOFT` dependency type to a current dinit version I had noticed that these return statements are missing.
(I would like to discuss bringing `SOFT` dependencies back in a discussion thread later).

If a request with an unknown dependency type is issued, the dinit will respond with BADREQ+ACK and add the invalid dependency to the service's dependency list (which I think is a bug that should be fixed).

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -932,6 +932,7 @@ bool control_conn_t::add_service_dep(bool do_enable, bool with_status_response)
         char badreqRep[] = { (char)cp_rply::BADREQ };
         if (!queue_packet(badreqRep, 1)) return false;
         bad_conn_close = true;
+        return true;
     }
     dependency_type dep_type = static_cast<dependency_type>(dep_type_int);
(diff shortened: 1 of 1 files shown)


davmac314 (approved):

davmac314:
This change looks good, thanks.

> (I would like to discuss bringing SOFT dependencies back in a discussion thread later).

You're welcome to do that if it needs further discussion, but note that in general you can get the same effect by introducing a stand-in dependency which depends on the relevant two dependencies.

Eg if you want service A with a soft "dependency" on B, you can instead introduce C (`type = internal`) which `depends-on: A` and `waits-for: B`, and then use C in place of A.

PR #583, opened 2026-09-05
Restore the signal mask before soft reboot

hholst80:
Soft reboot leaves signals blocked in services. On Ironforge, this stopped seatd from handling SIGUSR1 and SIGUSR2, which broke VT switching.

Dinit blocks signals for its own event loop. When it re-executes itself, the new instance saves that blocked mask as `orig_signal_mask` and passes it to services.

Restore the startup mask before `execv`. If `execv` fails, restore the internal mask and keep the error for logging. The shutdown hook still runs with the same mask as before.

Tested on Linux:

- `make check` and `make check-igr` passed (30 integration tests).
- Ran dinit as PID 1 in a separate user/PID namespace and chroot, with a harmless replacement shutdown helper. Unpatched 0.22.1 left SIGUSR1 and SIGUSR2 blocked after soft reboot. Patched 0.22.1 and patched master handled both signals through three soft reboots. SIGWINCH, deliberately blocked at startup, stayed blocked.
- Removed the test instance's dinit executable before soft reboot. The fallback restored the internal mask and logged the original ENOENT error.

Prepared with OpenAI Codex at hholst80's request. The contribution guide generally excludes AI-produced code, so this PR is a draft with that assistance disclosed.

### DIFF
--- a/src/dinit.cc
+++ b/src/dinit.cc
@@ -909,9 +909,19 @@ static void do_soft_reboot(char **argv) noexcept
         }
     }
 
+    // Preserve the startup mask across exec, rather than passing our internal
+    // blocked mask to the new instance and the services it starts.
+    sigset_t saved_signal_mask;
+    if (sigprocmask(SIG_SETMASK, &orig_signal_mask, &saved_signal_mask) == -1) {
+        log(loglevel_t::ERROR, "Could not restore signal mask for soft reboot: ", strerror(errno));
+        return;
+    }
+
     // Re-exec the dinit process.
     execv(dinit_exec, argv);
-    log(loglevel_t::ERROR, error_exec_dinit, strerror(errno));
+    int exec_errno = errno;
+    sigprocmask(SIG_SETMASK, &saved_signal_mask, nullptr);
+    log(loglevel_t::ERROR, error_exec_dinit, strerror(exec_errno));
 }
 
 // Callback for control socket


hholst80:
I did not see another way to get VT switching working after a soft reboot unless I carry this patch. Care to pick it up?

davmac314:
>  The contribution guide generally excludes AI-produced code

Yes, as you've noted. Thanks for being honest about using AI. Unfortunately I can't accept the PR and won't look at the code changes in it (to avoid any perception of contamination).

In general, it would probably be better for cases like this if you just open an issue describing the problem - it should be a simple enough fix and if it's a genuine issue then I am happy to make such a change. (No need to do that now, I can make this change regardless).

FWIW nothing should really be relying on signals being unmasked when they execute, since signal masks are always inherited from the parent process and there is no guarantee as to what signals are unmasked. I would personally therefore classify this also as a seatd bug; it could also be fixed there. I do agree though that it would be better if Dinit was consistent between ordinary boot and soft reboot.

hholst80:
The issue is created here: https://github.com/davmac314/dinit/blob/24fb3d859283758169c528f70841aab813e3e3f5/src/dinit.cc#L526-L540

In a soft-reboot scenario dinit exec dinit itself using this restricted set. It thereafter loose (apparently) a way to switch VT terminals which relies on signals. I think the reason is that the forked seatd in the soft-reboot will be more restricted because the signalling will be blocked to seatd.

seatd is one of these things I have questioned several times myself. "Do we really need this? can't we get rid of it?", but it is still lingering here so I guess I did not find an easy way to get rid of it.

---

• seatd uses:

  - SIGUSR1 — the kernel tells seatd to release the current VT.
  - SIGUSR2 — the kernel tells seatd it has acquired the new VT.

  Strictly speaking, seatd does not send these signals to initiate the switch. It requests the target
  VT using ioctl(VT_ACTIVATE, vt), then the kernel sends the release/acquire signals.

  Pointers:

  - Configure SIGUSR1 and SIGUSR2
    (https://github.com/kennylevinsen/seatd/blob/master/common/terminal.c#L173-L187)

  - Handle both signals (https://github.com/kennylevinsen/seatd/blob/master/seatd/server.c#L34-L35)
  - Request the switch with VT_ACTIVATE
    (https://github.com/kennylevinsen/seatd/blob/master/common/terminal.c#L204-L216)

  That is why blocking either signal can leave VT switching stuck.

